Issue #6, opened 2023-08-07
Feature Request: Improved Secrets redaction


ElijahSwiftIBM:
IRRSMO00 has an optional feature already to redact secret fields, and customers may also have fields beyond passwords and password phrases that require redaction. Both of these are features we should implement for additional security with pyRACF

ElijahSwiftIBM:
Created pull request #9 to address this concern

lcarcaramo:
PR9 merged.

PR #9, opened 2023-08-11
Secrets redaction

ElijahSwiftIBM:
### :bulb: Issue Reference

**Issue:**  #6 

Leverages existing IRRSMO00 functionality to better redact passwords, password phrases and other fields. Also better cleans up references to secrets in the python code, and allows for adding to the list of redacted secrets (password and password phrase). Also adds UserAdmin.set_passphrase function.

Changed options to always include option 8 in IRRSMO00. Drop references to parameter lists and strings containing input information when no longer used. Redact using regex rather than character replacement.

Added a test case and made adjustments to existing test case suite for password and password phrase redaction


pyracf/access/access_admin.py
```diff
@@ -40,13 +40,15 @@ def __init__(
         generate_requests_only: bool = False,
         add_field_data: Union[dict, None] = None,
         overwrite_field_data: Union[dict, None] = None,
+        add_more_secrets: Union[dict, None] = None,
```
lcarcaramo, line 43: I would call this field `additional_secret_traits`.

pyracf/common/logger.py
```diff
@@ -93,33 +94,63 @@ def log_debug(self, header_message: str, message: str) -> None:
         )
         print(f"{header}\n{message}")
 
-    def redact_strings(
+    def __redact_dict(
```
lcarcaramo, line 97: I would change `__redact_dict` to `__redact_request_dictionary` dictionary since this only redacts values in a request dictionary.

pyracf/common/security_admin.py
```diff
@@ -55,6 +61,7 @@ def __init__(
         generate_requests_only: bool = False,
         add_field_data: Union[dict, None] = None,
         overwrite_field_data: Union[dict, None] = None,
+        add_more_secrets: Union[dict, None] = None,
```
lcarcaramo, line 64: I would call this field `additional_secret_traits`.

pyracf/common/security_admin.py
```diff
@@ -84,6 +93,12 @@ def __overwrite_field_data(self, field_data: dict):
         """Overwrite field data in valid segment traits dictionary"""
         self._valid_segment_traits = field_data
 
+    def __add_more_secrets(self, new_secrets: dict):
```
lcarcaramo, line 96: I would call this field `__add_additional_secret_traits`.

pyracf/data_set/data_set_admin.py
```diff
@@ -60,13 +60,15 @@ def __init__(
         generate_requests_only: bool = False,
         add_field_data: Union[dict, None] = None,
         overwrite_field_data: Union[dict, None] = None,
+        add_more_secrets: Union[dict, None] = None,
```
lcarcaramo, line 63: I would call this field `additional_secret_traits`.

pyracf/resource/resource_admin.py
```diff
@@ -175,13 +175,15 @@ def __init__(
         generate_requests_only: bool = False,
         add_field_data: Union[dict, None] = None,
         overwrite_field_data: Union[dict, None] = None,
+        add_more_secrets: Union[dict, None] = None,
```
lcarcaramo, line 178: I would call this field `additional_secret_traits`.

pyracf/user/user_admin.py
```diff
@@ -209,13 +209,15 @@ def __init__(
         generate_requests_only: bool = False,
         add_field_data: Union[dict, None] = None,
         overwrite_field_data: Union[dict, None] = None,
+        add_more_secrets: Union[dict, None] = None,
```
lcarcaramo, line 212: I would call this field `additional_secret_traits`.

pyracf/common/logger.py
```diff
@@ -93,33 +94,63 @@ def log_debug(self, header_message: str, message: str) -> None:
...
+        for xml_key in secret_traits.values():
+            match = re.search(r"\<{}+[^>]*\>".format(xml_key), xml_string)
+            if not match:
+                continue
+            xml_string = self.__redact_string(xml_string, match.end(), "</")
+        if is_bytes:
+            xml_string = xml_string.encode("utf-8")
+        return xml_string
+
+    def redact_result_xml(
```
lcarcaramo, line 142: Do we even need `redact_result_xml()` anymore if we are now depending on the functionality built into IRRSMO0 to redact results now?

pyracf/common/security_admin.py
```diff
@@ -219,6 +211,15 @@ def _clear_state(self) -> None:
         self._segment_traits = {}
         self._trait_map = {}
 
+    def _reset_state(self) -> None:
+        """Reset state once requests are done."""
+        del self._segment_traits
+        del self._trait_map
+        del self.__preserved_segment_traits
```
lcarcaramo, line 218: I am not sure if deleting these values is necessary. I think it should be ok to just reset them to empty dictionary.

lcarcaramo: I am guessing that the del might be a security thing maybe. Is the purpose to remove this data completely from memory instead of just changing the references to ensure that no sensitive data is accessible in memory after the request is complete?

lcarcaramo:
One other thing to note is that now that IRRSMO00 handles redacting sensitive data from results, the password and passphrase redaction tests are actually now redundant. We only really need to verify that all secrets in debug logging are redacted and that request xml returned when using `generate_requests_only` is redacted. One thing that might be good to validate though is the clear state functionality if possible. Either way, I think we can get rid of the password and passphrase redaction tests in test_user_result_parser.py along with corresponding constants and xml samples since they don't really test anything anymore. Again, also note that anything that redacts results in `logger.py` is also now unnecessary since irrsmo00 handles redacting results for us as I have noted in feedback I provided earlier.

pyracf/access/access_admin.py
ElijahSwiftIBM: agree; that's better. change made

pyracf/common/logger.py
ElijahSwiftIBM: Agree, this is more descriptive and accurate. Change made

pyracf/common/logger.py
```diff
@@ -48,24 +49,24 @@ def log_dictionary(
...
         header_message: str,
         dictionary: dict,
-        redact_strings: List[str] = [],
+        secret_traits: dict = {},
     ) -> None:
         """JSONify and colorize a dictionary and log it to the console."""
-        dictionary_json = json.dumps(dictionary, indent=2)
-        dictionary_json = self.redact_strings(dictionary_json, redact_strings)
+        redacted_dict = self.__redact_dict(dictionary, secret_traits)
+        dictionary_json = json.dumps(redacted_dict, indent=2)
```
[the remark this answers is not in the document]
ElijahSwiftIBM: Agree; this is a good check and can save time and increase legibility of code.

pyracf/common/security_admin.py
```diff
@@ -108,30 +123,7 @@ def _extract_and_check_result(
...
-            irrsmo00_options=irrsmo00_options,
-            redact_strings=[redact_password, redact_passphrase],
-        )
-
-    def __make_request_and_redact_secrets(
-        self,
-        security_request: SecurityRequest,
-        irrsmo00_options: int = 1,
-        redact_strings: List[str] = [],
+        irrsmo00_options: int = 9,
```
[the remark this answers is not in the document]
ElijahSwiftIBM: I agree. This is more secure. change is implemented as suggested.

pyracf/common/security_admin.py
ElijahSwiftIBM: per above, agree.

pyracf/common/security_admin.py
ElijahSwiftIBM: agree, this is clearer and more descriptive. Added.

pyracf/connection/connection_admin.py
```diff
@@ -36,13 +36,15 @@ def __init__(
         generate_requests_only: bool = False,
         add_field_data: Union[dict, None] = None,
         overwrite_field_data: Union[dict, None] = None,
+        add_more_secrets: Union[dict, None] = None,
```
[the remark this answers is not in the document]
ElijahSwiftIBM: per above, agree.

pyracf/data_set/data_set_admin.py
ElijahSwiftIBM: per above, agree

pyracf/group/group_admin.py
```diff
@@ -45,13 +45,15 @@ def __init__(
         generate_requests_only: bool = False,
         add_field_data: Union[dict, None] = None,
         overwrite_field_data: Union[dict, None] = None,
+        add_more_secrets: Union[dict, None] = None,
```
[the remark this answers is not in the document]
ElijahSwiftIBM: per above, agree.

pyracf/resource/resource_admin.py
ElijahSwiftIBM: per above, agree.

pyracf/setropts/setropts_admin.py
```diff
@@ -103,13 +103,15 @@ def __init__(
         generate_requests_only: bool = False,
         add_field_data: Union[dict, None] = None,
         overwrite_field_data: Union[dict, None] = None,
+        add_more_secrets: Union[dict, None] = None,
```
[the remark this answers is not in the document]
ElijahSwiftIBM: per above, agree.

pyracf/user/user_admin.py
ElijahSwiftIBM: per above, agree.

pyracf/common/logger.py
ElijahSwiftIBM: Yes. First, this allows use of add_additional_secret traits to define traits as secret that IRRSMO00 may not additionally define. Additionally, IRRSMO00's method of redaction may not be sufficient for all cases, and this builds upon that redaction in a meaningful way.

pyracf/common/security_admin.py
ElijahSwiftIBM: That was the intention. With python's memory management, I'm not certain that it's really any better than setting them to empty, but it seemed like it was the more security-conscious thing to at least try to do.

pyracf/common/security_admin.py
```diff
@@ -141,28 +133,28 @@ def __make_request_and_redact_secrets(
...
                 security_request.dump_request_xml(encoding="utf-8"),
-                redact_strings=redact_strings,
+                secret_traits=self._secret_traits,
             )
         result_xml = self.__irrsmo00.call_racf(
             security_request.dump_request_xml(), irrsmo00_options
         )
+        result_xml = self.__logger.redact_result_xml(result_xml, self._secret_traits)
+        self._reset_state()
+        security_request.reset_request_xml()
```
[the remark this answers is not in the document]
ElijahSwiftIBM: Yes, this was the intention. Completely resetting the state of the request objects through deleting as much of their information as possible once the inputs have been fully utilized.

ElijahSwiftIBM:
> One other thing to note is that now that IRRSMO00 handles redacting sensitive data from results, the password and passphrase redaction tests are actually now redundant. We only really need to verify that all secrets in debug logging are redacted and that request xml returned when using `generate_requests_only` is redacted. One thing that might be good to validate though is the clear state functionality if possible. Either way, I think we can get rid of the password and passphrase redaction tests in test_user_result_parser.py along with corresponding constants and xml samples since they don't really test anything anymore. Again, also note that anything that redacts results in `logger.py` is also now unnecessary since irrsmo00 handles redacting results for us as I have noted in feedback I provided earlier.

I would like to keep the redaction tests as they serve as validation of the redaction of the result xml. While we now rely on IRRSMO00's redaction somewhat, ours is a bit more comprehensive, and testing us allows to ensure that this functionality persists with future releases.

pyracf/common/security_admin.py
```diff
@@ -84,6 +93,12 @@ def __overwrite_field_data(self, field_data: dict):
         """Overwrite field data in valid segment traits dictionary"""
         self._valid_segment_traits = field_data
 
+    def __add_additional_secret_traits(self, new_secrets: dict):
```
lcarcaramo, line 96: Change `new_secrets` to `additional_secret_traits` for consistency and because `new_secrets` implies that we are overwriting the existing `_secret_traits` dictionary when in reality we are just appending additional secrets to `_secret_traits`.

pyracf/common/security_admin.py
```diff
@@ -46,6 +46,12 @@ class SecurityAdmin:
         "base:gaudupdt": "racf:gaudupdt",
         "base:generic": "racf:generic",
     }
+
+    _secret_traits = {
```
lcarcaramo, line 50: I don't think that `_secret_traits` gets used anywhere else besides `security_admin.py`. In that case, it should start with `__` so that it can't be used anywhere outside `security_admin.py` including child classes. So, this field should be changed to `__secret_traits`.

pyracf/common/security_admin.py
```diff
@@ -219,6 +211,15 @@ def _clear_state(self) -> None:
...
         self._trait_map = {}
 
+    def _reset_state(self) -> None:
+        """Reset state once requests are done."""
+        del self._segment_traits
+        del self._trait_map
+        del self.__preserved_segment_traits
+        self._segment_traits = {}
+        self._trait_map = {}
+        self.__preserved_segment_traits = {}
```
lcarcaramo, line 221: It looks like we now have two different versions of this function in the form of `_clear_state` and `_reset_state` which accomplish similar things. I think that since all requests go through the `_make_request` function one at a time, we can just have one cleanup function call there. So, I think we can simplify things by just having the `_clear_state` function do everything. There are also some calls to `_clear_state` in some child classes, but I don't think this is necessary if we just make one call in `_make_request` to clean everything up at the end of the request, so we can change the name of the function to `__clear_state` to keep this private to `security_admin.py`.

```python
    def __clear_state(self) -> None:
        """Clear state for new request."""
        self._segment_traits = {}
        self._trait_map = {}
        self.__preserved_segment_traits = {}
        security_request.reset_request_xml()
```

pyracf/common/logger.py
```diff
@@ -93,33 +95,63 @@ def log_debug(self, header_message: str, message: str) -> None:
...
+        secret_traits: dict,
+    ) -> dict:
+        """Redact a list of specified secret traits in a dictionary"""
+        for seg_trait in secret_traits:
+            segment = seg_trait.split(":")[0]
+            try:
+                if not dictionary.get(segment, {}).get(seg_trait, {}).get("value", ""):
+                    continue
+                dictionary[segment][seg_trait]["value"] = "********"
+            except AttributeError:
```
lcarcaramo, line 110: I am guessing that the exception handler here is for line `109`

```python
               dictionary[segment][seg_trait]["value"] = "********"
```

In that case, you should be able to limit the scope of that exception handler to just that line. Also, it is better practice to ignore an exception by using `pass`. I think you should only use `continue` if there is code after that try except block that you want to skip. Also, changing `dictionary.get(segment, {}).get(seg_trait, {}).get("value", "")` to `dictionary.get(segment, {}).get(seg_trait, {}).get("value", None)` is clearer in my opinion.

```python
    ) -> dict:
        """Redact a list of specified secret traits in a dictionary"""
        for seg_trait in secret_traits:
            segment = seg_trait.split(":")[0]
            if not dictionary.get(segment, {}).get(seg_trait, {}).get("value", None):
                continue
            try:
                dictionary[segment][seg_trait]["value"] = "********"
            except AttributeError:
                pass
        return dictionary
```

pyracf/common/irrsmo00.py
```diff
@@ -19,8 +19,9 @@ def __init__(self) -> None:
         # Initialize size of output buffer
         self.buffer_size = 100000
 
-    def call_racf(self, request_xml: bytes, options: int = 1) -> str:
+    def call_racf(self, request_xml: bytes, precheck: bool = False) -> str:
         """Make request to call_irrsmo00 from pyracf_backend Python extension."""
```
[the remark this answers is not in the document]
ElijahSwiftIBM: Agree, change made.

pyracf/common/security_admin.py
ElijahSwiftIBM: Confirmed that `_secret_traits` is only used in `security_admin.py`. Change made.

pyracf/common/security_admin.py
```diff
@@ -141,28 +133,28 @@ def __make_request_and_redact_secrets(
             self.__logger.log_dictionary(
```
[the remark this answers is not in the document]
ElijahSwiftIBM: Agree, changed to:

```python
    def _make_request(
        self,
        security_request: SecurityRequest,
        irrsmo00_precheck: bool = False,
    ) -> Union[dict, bytes]:
        """
        Make request to IRRSMO00.
        Note: Secrets are redacted from all data returned to the user and log messages.
        """
```

pyracf/common/logger.py
```diff
@@ -93,33 +95,63 @@ def log_debug(self, header_message: str, message: str) -> None:
...
+
+    def __redact_string(
+        self,
+        input_string: Union[str, bytes],
+        start_ind: int,
+        end_pattern: str,
+    ):
+        pre_keyword = input_string[:start_ind]
+        post_keyword = end_pattern.join(input_string[start_ind:].split(end_pattern)[1:])
+        return pre_keyword + "********" + end_pattern + post_keyword
```
lcarcaramo, line 122: Can you add a docstring to this function that explains it's purpose? Just explain how it is used to either redact an xml field that contains secrets from request XML or redact secrets in TSO messages in result XML.

pyracf/common/logger.py
```diff
@@ -93,33 +95,63 @@ def log_debug(self, header_message: str, message: str) -> None:
...
-        """Redact a string or sequence of byte in a bytes object."""
-        if not redact_string:
-            return string
-        redacted_string = "********"
-        if isinstance(redact_string, bytes):
-            redacted_string = b"********"
-        return string.replace(redact_string, redacted_string)
+        for xml_key in secret_traits.values():
+            racf_key = xml_key.split(":")[1] if ":" in xml_key else xml_key
+            match = re.search(r"{} +\(".format(racf_key.upper()), xml_string)
```
lcarcaramo, line 150: Does the uppercase version of racf_key always correspond to what shows up in the returned TSO commands? It seems like it does, but I am just double checking. 

Also, I think using this syntax for your regexes is a little cleaner:

```python
            match = re.search(f"\<{xml_key}+[^>]*\>", xml_string)
```

```python
            match = re.search(f"{racf_key.upper()} +\(", xml_string)
```

I tested this out and I believe this syntax is functionally the same. I am not a fan of using `.format` because it add more clutter in my opinion.

pyracf/common/logger.py
```diff
@@ -93,33 +95,63 @@ def log_debug(self, header_message: str, message: str) -> None:
...
+            xml_string = self.__redact_string(xml_string, match.end(), "</")
+        if is_bytes:
+            xml_string = xml_string.encode("utf-8")
+        return xml_string
+
+    def redact_result_xml(
+        self,
+        xml_string: str,
+        secret_traits: dict,
     ) -> str:
```
lcarcaramo, line 147: Can you add a docstring here to explain how this function works. A brief explanation with an example of what the input and output looks like should be sufficient. The inclusion of the regex makes this function more esoteric and less intuitive so I think this would be helpful.

pyracf/common/logger.py
```diff
@@ -93,33 +95,63 @@ def log_debug(self, header_message: str, message: str) -> None:
...
+        post_keyword = end_pattern.join(input_string[start_ind:].split(end_pattern)[1:])
+        return pre_keyword + "********" + end_pattern + post_keyword
 
-    def redact_string(
-        self, string: Union[str, bytes], redact_string: Union[str, bytes]
+    def redact_request_xml(
+        self,
+        xml_string: Union[str, bytes],
+        secret_traits: dict,
+    ) -> Union[str, bytes]:
```
lcarcaramo, line 128: Can you add a docstring here to explain how this function works. A brief explanation with an example of what the input and output looks like should be sufficient. The inclusion of the regex makes this function more esoteric and less intuitive so I think this would be helpful.

tests/user/test_user_added_secret_redact.py
```diff
@@ -0,0 +1,74 @@
...
+import tests.user.test_user_constants as TestUserConstants
+from pyracf import SecurityRequestError, UserAdmin
+from pyracf.common.irrsmo00 import IRRSMO00
+
+# Resolves F401
+__init__
+
+
+@patch("pyracf.common.irrsmo00.IRRSMO00.call_racf")
+class TestUserAddedSecretRedact(unittest.TestCase):
```
lcarcaramo, line 20: Is there a reason why we shouldn't be doing these tests in `TestUserDebugLogging`? Also, referering to these tests as testing "redact additional secrets" instead of "added secrets redact" is more clear to me.

pyracf/common/security_admin.py
```diff
@@ -68,6 +75,8 @@ def __init__(
             self.__add_field_data(add_field_data)
         if overwrite_field_data is not None:
             self.__overwrite_field_data(overwrite_field_data)
+        if additional_secret_traits is not None:
+            self.__add_additional_secret_traits(additional_secret_traits)
```
lcarcaramo, line 79: I think to reduce the likelihood of user error it makes sens for `additional_secret_traits` to just be a list. Then the user only needs to worry about the pyRACF trait name.

So instead of defining `additional_secret_traits` like this:

```python
user_admin = UserAdmin(additional_secret_traits={"omvs:uid": "uid"})
```

they can instead by added like this:

```python
user_admin = UserAdmin(additional_secret_traits=["omvs:uid"])
```

Then `__add_additional_secret_traits` would just need to lookup the IRRSMO00 key in `_valid_segment_traits`.

In short, do this to make the user's life easier.

pyracf/common/logger.py
ElijahSwiftIBM: Actually, the attribute error had been for the "if" statement. In testing (I think for extract requests), I was hitting Attribute Errors for requests since their segments dictionary leads to bools not more dictionaries. The existing solution was clunky, and should have been better optimized. I replaced it with the following and don't see the error in unit testing.

```python
    ) -> dict:
        """Redact a list of specified secret traits in a dictionary"""
        for seg_trait in secret_traits:
            segment = seg_trait.split(":")[0]
            if not isinstance(dictionary.get(segment, {}).get(seg_trait, {}),dict):
                return dictionary
            if not dictionary.get(segment, {}).get(seg_trait, {}).get("value", ""):
                continue
            dictionary[segment][seg_trait]["value"] = "********"
        return dictionary
```

pyracf/common/logger.py
ElijahSwiftIBM: Added the following docstring. It is a bit generic, but technically representative of the function's purpose.

````python
"""Redacts characters in a string between a starting index and ending pattern"""
```

pyracf/common/logger.py
ElijahSwiftIBM: Added the following docstring.

```python
"""Redacts a list of specific secret traits in a result xml string"""

File: pyracf/common/security_request.py [hunk 28]
```diff
@@ -92,3 +92,15 @@ def dump_request_xml(self, encoding: str = "cp1047") -> bytes:
...
+        """Delete the security request xml string"""
+        del self.__racf_request
+        self.__racf_request = XMLBuilder.Element("securityrequest")
+        self.__racf_request.attrib = {
+            "xmlns": "http://www.ibm.com/systems/zos/saf",
+            "xmlns:racf": "http://www.ibm.com/systems/zos/racf",
+        }
+        self._security_definition = XMLBuilder.SubElement(
+            self.__racf_request, "undefined"
+        )
```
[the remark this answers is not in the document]
ElijahSwiftIBM: This does make more sense and does not introduce new issues. I have added this `del` to `_make_request` in `security_admin.py` and deleted the function intended to reset the request object.

tests/user/test_user_added_secret_redact.py
ElijahSwiftIBM: The tests have to be separate and other user tests will break if this new test is not the last one run. Since it hinges on redacting a field used elsewhere in testing (`omvs:uid`), defining the object with this additional secret breaks other tests as there is no supported mechanism currently in place to remove a secret. I concede that it is entirely possible there is a better way to force this test to have a separate instance of UserAdmin that would not impact other tests, but my attempts to do so did not work.

I did change the names as indicated.

pyracf/common/security_admin.py
ElijahSwiftIBM: Agree, this makes more sense. With some tweaks, this brings the `__add_additional_secret_traits` function to the following.

```python
    def __add_additional_secret_traits(self, new_secrets: list):
        """Add additional fields to be redacted in logger output"""
        for secret in new_secrets:
            if secret in self._secret_traits:
                continue
            if ':' not in secret:
                continue
            segment = secret.split(':')[0]
            if segment not in self._valid_segment_traits:
                continue
            if secret not in self._valid_segment_traits[segment]:
                continue
            self._secret_traits[secret] = self._valid_segment_traits[segment][secret]
```

pyracf/common/logger.py
ElijahSwiftIBM: Changed again to `rf` strings to avoid flake8 errors. Still passes all unit tests.

pyracf/common/irrsmo00.py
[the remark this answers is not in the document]
lcarcaramo: You must have forgot to commit this change. I am not seeing the change.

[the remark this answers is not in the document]
lcarcaramo: I see the change now.

pyracf/common/security_admin.py
lcarcaramo: I am not seeing this change in the code.

pyracf/access/access_admin.py
```diff
@@ -40,13 +40,15 @@ def __init__(
         generate_requests_only: bool = False,
         add_field_data: Union[dict, None] = None,
         overwrite_field_data: Union[dict, None] = None,
+        additional_secret_traits: Union[list, None] = None,
```
lcarcaramo, line 43:
```python
        additional_secret_traits: Union[List[str], None] = None,
```

pyracf/common/security_admin.py
```diff
@@ -55,6 +61,7 @@ def __init__(
         generate_requests_only: bool = False,
         add_field_data: Union[dict, None] = None,
         overwrite_field_data: Union[dict, None] = None,
+        additional_secret_traits: Union[list, None] = None,
```
lcarcaramo, line 64:
```python
        additional_secret_traits: Union[List[str], None] = None,
```

pyracf/connection/connection_admin.py
```diff
@@ -36,13 +36,15 @@ def __init__(
         generate_requests_only: bool = False,
         add_field_data: Union[dict, None] = None,
         overwrite_field_data: Union[dict, None] = None,
+        additional_secret_traits: Union[list, None] = None,
```
lcarcaramo, line 39:
```python
        additional_secret_traits: Union[List[str], None] = None,
```

pyracf/data_set/data_set_admin.py
```diff
@@ -60,13 +60,15 @@ def __init__(
         generate_requests_only: bool = False,
         add_field_data: Union[dict, None] = None,
         overwrite_field_data: Union[dict, None] = None,
+        additional_secret_traits: Union[list, None] = None,
```
lcarcaramo, line 63:
```python
        additional_secret_traits: Union[List[str], None] = None,
```

pyracf/group/group_admin.py
```diff
@@ -45,13 +45,15 @@ def __init__(
         generate_requests_only: bool = False,
         add_field_data: Union[dict, None] = None,
         overwrite_field_data: Union[dict, None] = None,
+        additional_secret_traits: Union[list, None] = None,
```
lcarcaramo, line 48:
```python
        additional_secret_traits: Union[List[str], None] = None,
```

pyracf/resource/resource_admin.py
```diff
@@ -175,13 +175,15 @@ def __init__(
         generate_requests_only: bool = False,
         add_field_data: Union[dict, None] = None,
         overwrite_field_data: Union[dict, None] = None,
+        additional_secret_traits: Union[list, None] = None,
```
lcarcaramo, line 178:
```python
        additional_secret_traits: Union[List[str], None] = None,
```

pyracf/user/user_admin.py
```diff
@@ -209,13 +209,15 @@ def __init__(
         generate_requests_only: bool = False,
         add_field_data: Union[dict, None] = None,
         overwrite_field_data: Union[dict, None] = None,
+        additional_secret_traits: Union[list, None] = None,
```
lcarcaramo, line 212:
```python
        additional_secret_traits: Union[List[str], None] = None,
```

pyracf/common/security_admin.py
```diff
@@ -248,7 +250,7 @@ def __validate_and_add_trait(
     def _build_bool_segment_dictionaries(self, segments: dict) -> None:
         """Build segment dictionaries for profile extract."""
         # Clear state for new request
-        self._clear_state()
+        self.__clear_state()
```
lcarcaramo, line 253: I don't think this is necessary since state is cleared at the end of `_make_request`.

pyracf/common/security_admin.py
```diff
@@ -258,7 +260,7 @@ def _build_bool_segment_dictionaries(self, segments: dict) -> None:
     def _build_segment_dictionaries(self, traits: dict) -> None:
         """Build segemnt dictionaries for each segment."""
         # Clear state for new request
-        self._clear_state()
+        self.__clear_state()
```
lcarcaramo, line 263: I don't think this is necessary since state is cleared at the end of `_make_request`.

pyracf/common/security_admin.py
```diff
@@ -286,7 +288,7 @@ def _add_traits_directly_to_request_xml_with_no_segments(
                     None, traits, self._trait_map, alter=alter
                 )
         # Clear state for subsequent requests
-        self._clear_state()
+        self.__clear_state()
```
lcarcaramo, line 291: I don't think this is necessary since state is cleared at the end of `_make_request`.

pyracf/resource/resource_admin.py
```diff
@@ -246,7 +248,7 @@ def extract(
 
     def delete(self, resource: str, class_name: str) -> Union[dict, bytes]:
         """Delete a general resource profile."""
-        self._clear_state()
+
```
lcarcaramo, line 251: Get rid of extraneous new line.

pyracf/user/user_admin.py
```diff
@@ -410,7 +424,7 @@ def extract(
 
     def delete(self, userid: str) -> Union[dict, bytes]:
         """Delete a user."""
-        self._clear_state()
+
```
lcarcaramo, line 427: Get rid of extraneous new line.

pyracf/access/access_admin.py
ElijahSwiftIBM: Agree, fixed in 8 files.

pyracf/connection/connection_admin.py
ElijahSwiftIBM: Agree, fixed in 8 files.

pyracf/data_set/data_set_admin.py
ElijahSwiftIBM: Agree, fixed in 8 files.

pyracf/group/group_admin.py
ElijahSwiftIBM: Agree, fixed in 8 files.

pyracf/resource/resource_admin.py
ElijahSwiftIBM: Agree, fixed in 8 files.

pyracf/setropts/setropts_admin.py
```diff
@@ -103,13 +103,15 @@ def __init__(
         generate_requests_only: bool = False,
         add_field_data: Union[dict, None] = None,
         overwrite_field_data: Union[dict, None] = None,
+        additional_secret_traits: Union[list, None] = None,
```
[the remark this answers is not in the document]
ElijahSwiftIBM: Agree, fixed in 8 files.

pyracf/user/user_admin.py
ElijahSwiftIBM: fixed

pyracf/resource/resource_admin.py
ElijahSwiftIBM: fixed

pyracf/group/group_admin.py
```diff
@@ -149,7 +151,7 @@ def extract(
 
     def delete(self, group: str) -> Union[dict, bytes]:
         """Delete a group."""
-        self._clear_state()
+
```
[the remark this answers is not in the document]
ElijahSwiftIBM: fixed

pyracf/common/logger.py
lcarcaramo: This doesn't seem quite right. You don't want to return prematurely. I think something like this should work:

```python
    ) -> dict:
        """Redact a list of specified secret traits in a dictionary"""
        for seg_trait in secret_traits:
            segment = seg_trait.split(":")[0]
            if dictionary.get(segment, {}).get(seg_trait, {}).get("value", None):
                continue
            dictionary[segment][seg_trait]["value"] = "********"
        return dictionary
```

If the value doesn't exist `continue` and don't try to redact anything, if you do find the value, redact it.

pyracf/common/security_admin.py
lcarcaramo: Skipping bad input is OK for now, but as we discussed, we will definitely want to raise an exception or something in a future update when bad input is given.

tests/user/test_user_added_secret_redact.py
lcarcaramo: I think I know why this is breaking your tests.

Notice how __secret_traits is initialized once here.

```python
class SecurityAdmin:
    """Base Class for RACF Administration Interface."""

    _valid_segment_traits = {}
    _common_base_traits_data_set_generic = {
        "base:aclcnt": "racf:aclcnt",
        "base:aclacnt": "racf:aclacnt",
        "base:aclacs": "racf:aclacs",
        "base:aclid": "racf:aclid",
        "base:acl2cnt": "racf:acl2cnt",
        "base:acl2acnt": "racf:acl2acnt",
        "base:acl2acs": "racf:acl2acs",
        "base:acl2cond": "racf:acl2cond",
        "base:acl2ent": "racf:acl2ent",
        "base:acl2id": "racf:acl2id",
        "base:acsaltr": "racf:acsaltr",
        "base:acscntl": "racf:acscntl",
        "base:acsread": "racf:acsread",
        "base:acsupdt": "racf:acsupdt",
        "base:all": "racf:all",
        "base:audaltr": "racf:audaltr",
        "base:audcntl": "racf:audcntl",
        "base:audnone": "racf:audnone",
        "base:audread": "racf:audread",
        "base:audupdt": "racf:audupdt",
        "base:authuser": "racf:authuser",
        "base:fvolume": "racf:fvolume",
        "base:gaudaltr": "racf:gaudaltr",
        "base:gaudcntl": "racf:gaudcntl",
        "base:gaudnone": "racf:gaudnone",
        "base:gaudread": "racf:gaudread",
        "base:gaudupdt": "racf:gaudupdt",
        "base:generic": "racf:generic",
    }

    __secret_traits = {
        "base:password": "racf:password",
        "base:passphrase": "racf:phrase",
    }

    __logger = Logger()

    def __init__(
        self,
        profile_type: str,
        debug: bool = False,
        generate_requests_only: bool = False,
        add_field_data: Union[dict, None] = None,
        overwrite_field_data: Union[dict, None] = None,
        additional_secret_traits: Union[List[str], None] = None,
    ) -> None:
        self.__irrsmo00 = IRRSMO00()
        self.__profile_type = profile_type
        self._segment_traits = {}
        # used to preserve segment traits for debug logging.
        self.__preserved_segment_traits = {}
        self._trait_map = {}
        self.__debug = debug
        self.__generate_requests_only = generate_requests_only
        if add_field_data is not None:
            self.__add_field_data(add_field_data)
        if overwrite_field_data is not None:
            self.__overwrite_field_data(overwrite_field_data)
        if additional_secret_traits is not None:
            self.__add_additional_secret_traits(additional_secret_traits)
```

I think moving the initialization of `__secret_traits` to the constructor would resolve the issue since __secret_traits would be rebuild every time you call the constructor.

It is also worth noting that the custom field data may actually lead to the same problem and that each child classes definition of `_valid_segment_traits` might be better defined as part of the constructor too.

pyracf/common/logger.py
```diff
@@ -93,33 +95,64 @@ def log_debug(self, header_message: str, message: str) -> None:
...
-                string = self.redact_string(
-                    string, redact_string=bytes(to_redact, "utf-8")
-                )
-            else:
-                string = self.redact_string(string, redact_string=to_redact)
-        return string
+        dictionary: dict,
+        secret_traits: dict,
+    ) -> dict:
+        """Redact a list of specified secret traits in a dictionary"""
```
lcarcaramo, line 103:
```
        """Redact a list of specified secret traits in a request dictionary"""
```

ElijahSwiftIBM: agree, fixed

pyracf/common/security_admin.py
ElijahSwiftIBM: Apologies, that change must have been lost when stashing changes between branches. With later changes this now becomes the following.

```python
    def __add_additional_secret_traits(self, additional_secret_traits: list):
        """Add additional fields to be redacted in logger output"""
        for secret in additional_secret_traits:
            if secret in self.__secret_traits:
                continue
            if ":" not in secret:
                continue
            segment = secret.split(":")[0]
            if segment not in self._valid_segment_traits:
                continue
            if secret not in self._valid_segment_traits[segment]:
                continue
            self.__secret_traits[secret] = self._valid_segment_traits[segment][secret]
```

pyracf/common/logger.py
ElijahSwiftIBM: The intent behind that first `if` statement is that if we meet the condition specified, it is because the traits dictionary is of the form of an extract, where it takes the form { 'segment': bool } instead of  { 'segment': { 'trait': {} } } as expected for other requests. we do not redact extract request dictionaries because there are no fields to redact, so there is no reason to continue.

With that in mind, your change did work with a slight correction, so the `get` behavior is not how I initially expected.

```python
    ) -> dict:
        """Redact a list of specified secret traits in a request dictionary"""
        for seg_trait in secret_traits:
            segment = seg_trait.split(":")[0]
            if not dictionary.get(segment, {}).get(seg_trait, {}).get("value", None):
                continue
            dictionary[segment][seg_trait]["value"] = "********"
        return dictionary
```

pyracf/common/logger.py
ElijahSwiftIBM: Changed to:

```python
        """Redact a list of specific secret traits in a request xml string or bytes object"""
        """Bases this on xml pattern <xml_key any-operation>xml_value</xml_key>"""
```

pyracf/common/logger.py
ElijahSwiftIBM: Added the following docstring:

```python
        """Redacts characters in a string between a starting index and ending pattern"""
        """Replaces the identified characters with ******** (regardless of length)"""
```

tests/user/test_user_added_secret_redact.py
ElijahSwiftIBM: That makes perfect sense. I have addressed this and the unit testing works perfectly normally, so I returned the tests to `test_user_debug_logging` where they still pass the test. I have also moved the `_valid_segment_traits` definitions into the class constructors as well, though I am surprised that this didn't cause similar problems in unit testing originally.

pyracf/common/security_admin.py
ElijahSwiftIBM: It still was necessary, since there was no state clearing for `__generate_requests_only`, which the unit tests originally ran on. I changed this code

```python
        if self.__generate_requests_only:
            return self.__logger.redact_request_xml(
                security_request.dump_request_xml(encoding="utf-8"),
                secret_traits=self.__secret_traits,
            )
```

to the following. 

```python
        if self.__generate_requests_only:
            request = self.__logger.redact_request_xml(
                security_request.dump_request_xml(encoding="utf-8"),
                secret_traits=self.__secret_traits,
            )
            self.__clear_state(clear_preserved_segment_traits=True)
            del security_request
            return request
```

This should happen anyway since we DO still want to clean up the inputs when we are generating a request only.

pyracf/common/security_admin.py
ElijahSwiftIBM: It still was necessary, since there was no state clearing for `__generate_requests_only`, which the unit tests originally ran on. I changed this code

```python
        if self.__generate_requests_only:
            return self.__logger.redact_request_xml(
                security_request.dump_request_xml(encoding="utf-8"),
                secret_traits=self.__secret_traits,
            )
```
to the following.

```python
        if self.__generate_requests_only:
            request = self.__logger.redact_request_xml(
                security_request.dump_request_xml(encoding="utf-8"),
                secret_traits=self.__secret_traits,
            )
            self.__clear_state(clear_preserved_segment_traits=True)
            del security_request
            return request
```

This should happen anyway since we DO still want to clean up the inputs when we are generating a request only.

pyracf/common/security_admin.py
ElijahSwiftIBM: It still was necessary, since there was no state clearing for `__generate_requests_only`, which the unit tests originally ran on. I changed this code

```python
        if self.__generate_requests_only:
            return self.__logger.redact_request_xml(
                security_request.dump_request_xml(encoding="utf-8"),
                secret_traits=self.__secret_traits,
            )
```
to the following.

```python
        if self.__generate_requests_only:
            request = self.__logger.redact_request_xml(
                security_request.dump_request_xml(encoding="utf-8"),
                secret_traits=self.__secret_traits,
            )
            self.__clear_state(clear_preserved_segment_traits=True)
            del security_request
            return request
```

This should happen anyway since we DO still want to clean up the inputs when we are generating a request only.

pyracf/common/logger.py
lcarcaramo: I agree. Even though this technically gets used with profile extract, this logic is cleaner and easier to follow. Even though this logic is pointless for profile extract, the amount of cycles that get wasted running through this loop is negligible. My only suggestion would be to maybe document this nuance in the docstring for this function so that there is no confusion about what happens with profile extract.

tests/user/test_user_added_secret_redact.py
lcarcaramo: So, I think the general rule of thumb is that class attributes that are static should be defined directly to the class.

```python
class MyClass():
    my_static_attribute = "Hello World!"
```

And class attributes that are dynamic and or specific to a an instance of a class should be initialized in the constructor.

```python
class MyClass():
    def __init__(self):
        my_dynamic_instance_specific_attribute = "Hello World!"
```

tests/user/test_user_debug_logging.py
```diff
@@ -184,6 +184,54 @@ def test_add_user_request_debug_log_passphrases_and_passwords_get_redacted_on_er
         self.assertNotIn(self.test_passphrase, error_log)
         self.assertNotIn(self.test_password, error_log)
 
+    def test_add_user_request_debug_log_added_secrets_get_redacted_on_success(
+        self,
+        call_racf_mock: Mock,
+    ):
+        self.user_admin_local = UserAdmin(
```
lcarcaramo, line 191: Since you are creating a local `UserAdmin` object, you do not need to define it as a class attribute. You can just do this:

```python
        user_admin = UserAdmin(
            debug=True, additional_secret_traits=["omvs:uid"]
        )
```

tests/user/test_user_result_parser.py
```diff
@@ -289,13 +303,13 @@ def test_user_admin_can_parse_extract_user_base_omvs_csdata_success_xml(
         self,
         call_racf_mock: Mock,
     ):
-        user_admin = UserAdmin(
+        user_admin_local = UserAdmin(
```
lcarcaramo, line 306: Again, I don't think it is necessary to call local variable local if they are already scoped. Leaving this as `user_admin` is fine.

pyracf/common/logger.py
```diff
@@ -93,33 +95,65 @@ def log_debug(self, header_message: str, message: str) -> None:
...
 
-    def redact_string(
-        self, string: Union[str, bytes], redact_string: Union[str, bytes]
+    def redact_request_xml(
+        self,
+        xml_string: Union[str, bytes],
+        secret_traits: dict,
+    ) -> Union[str, bytes]:
+        """Redact a list of specific secret traits in a request xml string or bytes object"""
+        """Bases this on xml pattern <xml_key any-operation>xml_value</xml_key>"""
```
lcarcaramo, line 129:
```python
        """
        Redact a list of specific secret traits in a request xml string or bytes object.
        Based the following xml pattern:
            '<xmltag attribute="any">xml value</xmltag>'
        """
```

pyracf/common/logger.py
```diff
@@ -93,33 +95,65 @@ def log_debug(self, header_message: str, message: str) -> None:
...
     ) -> str:
-        """Redact a string or sequence of byte in a bytes object."""
-        if not redact_string:
-            return string
-        redacted_string = "********"
-        if isinstance(redact_string, bytes):
-            redacted_string = b"********"
-        return string.replace(redact_string, redacted_string)
+        """Redacts a list of specific secret traits in a result xml string"""
+        """Bases this on RACF command pattern RACF_KEY( value ) """
```
lcarcaramo, line 149:
```python
        """
        Redacts a list of specific secret traits in a result xml string.
        Based on the following RACF command pattern:
            'TRAIT (value)'
        """
```

Also, I am guessing that this ignores white space since there is always a lot of white space in the RACF commands returned by IRRSMO00. In this case, I think it would be helpful to add a comment about this in the docstring.

pyracf/common/logger.py
```diff
@@ -93,33 +95,65 @@ def log_debug(self, header_message: str, message: str) -> None:
...
+        return dictionary
+
+    def __redact_string(
+        self,
+        input_string: Union[str, bytes],
+        start_ind: int,
+        end_pattern: str,
+    ):
+        """Redacts characters in a string between a starting index and ending pattern"""
+        """Replaces the identified characters with ******** (regardless of length)"""
```
lcarcaramo, line 118:
```
        """
        Redacts characters in a string between a starting index and ending pattern.
        Replaces the identified characters with '********' regardless of the original length.
        """
```

tests/user/test_user_constants.py
```diff
@@ -198,6 +199,9 @@ def get_sample(sample_file: str) -> Union[str, bytes]:
 TEST_ADD_USER_SUCCESS_LOG = get_sample("add_user_success.log")
 TEST_ADD_USER_ERROR_LOG = get_sample("add_user_error.log")
 
+TEST_ADD_USER_ADDED_SECRET_SUCCESS_LOG = get_sample("add_user_added_secret_success.log")
+TEST_ADD_USER_ADDED_SECRET_ERROR_LOG = get_sample("add_user_added_secret_error.log")
```
lcarcaramo, line 203: Can we use the naming convention "additional secret added" instead of "added secret" for these tests for clarity purposes?

tests/user/test_user_result_parser.py
```diff
@@ -157,7 +157,9 @@ def test_user_admin_password_redacted_add_user_success_xml(
             result,
             TestUserConstants.TEST_ADD_USER_PASSWORD_RESULT_SUCCESS_DICTIONARY,
         )
-        self.assertNotIn(self.test_password, str(result))
+        result_str = str(result)
+        self.assertNotIn(self.test_password, result_str)
+        self.assertNotIn("(" + " " * len(self.test_password) + ")", result_str)
```
lcarcaramo, line 162: Does it make sense to have a test where a password of "password" is used to verify that the xml tag `<racf:password>` does not get redacted here?

pyracf/common/logger.py
```diff
@@ -93,33 +95,65 @@ def log_debug(self, header_message: str, message: str) -> None:
...
-                    string, redact_string=bytes(to_redact, "utf-8")
-                )
-            else:
-                string = self.redact_string(string, redact_string=to_redact)
-        return string
+        dictionary: dict,
+        secret_traits: dict,
+    ) -> dict:
+        """Redact a list of specified secret traits in a request dictionary"""
+        for seg_trait in secret_traits:
```
lcarcaramo, line 104: This is small, but I would just call this variable `trait` instead of `seg_trait`.

tests/user/test_user_debug_logging.py
ElijahSwiftIBM: Good point, change made.

tests/user/test_user_debug_logging.py
```diff
@@ -184,6 +184,54 @@ def test_add_user_request_debug_log_passphrases_and_passwords_get_redacted_on_er
         self.assertNotIn(self.test_passphrase, error_log)
         self.assertNotIn(self.test_password, error_log)
 
+    def test_add_user_request_debug_log_added_secrets_get_redacted_on_success(
```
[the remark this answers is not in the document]
ElijahSwiftIBM: Makes sense, also added the following title for other tests here:


```python
    # Secrets Redaction

```

tests/user/test_user_request_builder.py
```diff
@@ -87,35 +87,35 @@ def test_user_admin_build_add_user_request_passphrase_and_password_redacted(
     # Custom Field Data
     # ============================================================================
     def test_user_admin_build_alter_request_alternate_segments(self):
-        user_admin = UserAdmin(
+        user_admin_local = UserAdmin(
```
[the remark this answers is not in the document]
ElijahSwiftIBM: Fair enough, I wanted to make it more obvious this was a local version of it with different attributes, but that should be clear just by the fact that this is built within the test. Change made.

tests/user/test_user_result_parser.py
ElijahSwiftIBM: Fair enough, I wanted to make it more obvious this was a local version of it with different attributes, but that should be clear just by the fact that this is built within the test. Change made.

pyracf/common/logger.py
ElijahSwiftIBM: That is a cleaner way of this level of docstring. I will make similar adjustments elsewhere.

ElijahSwiftIBM: Based on below comment on whitespace, I changed this to.

```python
        """
        Redact a list of specific secret traits in a request xml string or bytes object.
        Based the following xml pattern:
            '<xmltag attribute="any">xml value</xmltag>'
        This function also accounts for varied amounts of variation in the xml attributes.
        """
```

pyracf/common/logger.py
ElijahSwiftIBM: Now changed to:

```python
        """
        Redacts a list of specific secret traits in a result xml string.
        Based on the following RACF command pattern:
            'TRAIT (value)'
        This function also accounts for varied amounts of whitespace in the pattern.
        """
```

pyracf/common/logger.py
ElijahSwiftIBM: Agree; changed.

tests/user/test_user_constants.py
ElijahSwiftIBM: Agree, nice catch, I had missed that in the other name changes.

tests/user/test_user_result_parser.py
ElijahSwiftIBM: Added additional test in `test_user_debug_logging` and `test_user_result_parser` for a `simple_password` (PASSWORD). These tests assert that the password does not appear in the request dictionary, request xml, or either of the results using character strings that would surround the password in these cases ( "PASSWORD" >PASSWORD< and (PASSWORD)), while asserting that the actual string of PASSWORD is present because of the XML tags and the result message from the RACF command.

pyracf/common/security_admin.py
lcarcaramo: This conversation is still out standing.

pyracf/common/logger.py
ElijahSwiftIBM: Changed docstring to:

```python
        """
        Redact a list of specified secret traits in a request dictionary.
        While this flow is technically used in extract functions, there are no traits to redact.
        """
```

tests/user/test_user_debug_logging.py
```diff
@@ -184,6 +188,98 @@ def test_add_user_request_debug_log_passphrases_and_passwords_get_redacted_on_er
...
+        self.assertEqual(error_log, TestUserConstants.TEST_ADD_USER_PASSWORD_ERROR_LOG)
+        self.assertNotIn("(" + self.simple_password + ")", error_log)
+        self.assertNotIn('"' + self.simple_password + '"', error_log)
+        self.assertNotIn(">" + self.simple_password + "<", error_log)
+        self.assertIn(self.simple_password, error_log)
+
+    # ============================================================================
+    # Add Additional Secrets
+    # ============================================================================
+    def test_add_user_request_debug_log_additional_secret_addeds_get_redacted_on_success(
```
lcarcaramo, line 239: typo

tests/user/test_user_debug_logging.py
```diff
@@ -184,6 +188,98 @@ def test_add_user_request_debug_log_passphrases_and_passwords_get_redacted_on_er
...
+        success_log = self.ansi_escape.sub("", stdout.getvalue())
+        self.assertEqual(
+            success_log,
+            TestUserConstants.TEST_ADD_USER_ADDITIONAL_SECRET_ADDED_SUCCESS_LOG,
+        )
+        self.assertNotIn(
+            TestUserConstants.TEST_ADD_USER_REQUEST_TRAITS["omvs:uid"], success_log
+        )
+
+    def test_add_user_request_debug_log_additional_secret_addeds_get_redacted_on_error(
```
lcarcaramo, line 260: typo

tests/user/test_user_request_builder.py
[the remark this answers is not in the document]
lcarcaramo: Also, you know the variable is class scope because you use `self.`. Anything that is defined as a global constant should be in all caps.

tests/user/test_user_debug_logging.py
ElijahSwiftIBM: fixed

pyracf/common/logger.py
lcarcaramo: My guess is that the last comment is saying that it accounts for any number of arbitrary xml attributes?

```python
        """
        Redact a list of specific secret traits in a request xml string or bytes object.
        Based the following xml pattern:
            '<xmltag attribute="any">xml value</xmltag>'
        This function also accounts for any number of arbitrary xml attributes.
        """
```

ElijahSwiftIBM: I didn't realize I put varied and variation in the comment. I wanted to leave it more general because even no attributes or improperly defined ones will be ignored. it really looks for <xmltag[ANY or NO TEXT HERE]>. I think what you wrote is more than clear enough, so I will fix that.

pyracf/common/security_admin.py
```diff
@@ -108,61 +132,39 @@ def _extract_and_check_result(
...
+                secret_traits=self.__secret_traits,
             )
+            self.__clear_state(security_request=security_request)
+            return request
         result_xml = self.__irrsmo00.call_racf(
-            security_request.dump_request_xml(), irrsmo00_options
+            security_request.dump_request_xml(), irrsmo00_precheck
         )
+        result_xml = self.__logger.redact_result_xml(result_xml, self.__secret_traits)
+        self.__clear_state(security_request=security_request)
```
lcarcaramo, line 163: `security_request` is a positional argument. It does not need to be specified as a keyword argument.

```python
        self.__clear_state(security_request)
```

pyracf/common/security_admin.py
```diff
@@ -108,61 +132,39 @@ def _extract_and_check_result(
...
+                secret_traits=self.__secret_traits,
             )
         if self.__generate_requests_only:
-            return self.__logger.redact_strings(
+            request = self.__logger.redact_request_xml(
                 security_request.dump_request_xml(encoding="utf-8"),
-                redact_strings=redact_strings,
+                secret_traits=self.__secret_traits,
             )
+            self.__clear_state(security_request=security_request)
```
lcarcaramo, line 157: `security_request` is a positional argument. It does not need to be specified as a keyword argument.

```python
            self.__clear_state(security_request=security_request)
```

pyracf/common/security_admin.py
```diff
@@ -108,61 +132,39 @@ def _extract_and_check_result(
...
+            return request
         result_xml = self.__irrsmo00.call_racf(
-            security_request.dump_request_xml(), irrsmo00_options
+            security_request.dump_request_xml(), irrsmo00_precheck
         )
+        result_xml = self.__logger.redact_result_xml(result_xml, self.__secret_traits)
```
lcarcaramo, line 162: I think this can all just be combined into one thing:

```python
        result_xml = self.__logger.redact_result_xml(
            self.__irrsmo00.call_racf(
                security_request.dump_request_xml(), irrsmo00_precheck
            ),
            self.__secret_traits,
        )
```

pyracf/common/security_admin.py
lcarcaramo: Looks good now.

tests/user/test_user_debug_logging.py
```diff
@@ -184,6 +188,98 @@ def test_add_user_request_debug_log_passphrases_and_passwords_get_redacted_on_er
...
+                traits=TestUserConstants.TEST_ADD_USER_REQUEST_TRAITS_PASSWORD_SIMPLE,
+            )
+        success_log = self.ansi_escape.sub("", stdout.getvalue())
+        self.assertEqual(
+            success_log, TestUserConstants.TEST_ADD_USER_PASSWORD_SUCCESS_LOG
+        )
+        self.assertNotIn("(" + self.simple_password + ")", success_log)
+        self.assertNotIn('"' + self.simple_password + '"', success_log)
+        self.assertNotIn(">" + self.simple_password + "<", success_log)
+        self.assertIn(self.simple_password, success_log)
```
lcarcaramo, line 211: I don't know if the `assertNotIn` and `assertIn` statements are necessary here. I think it is enough to just check that the log file still looks like the sample even if you use a password that matches other things in the data.

ElijahSwiftIBM: Fair enough, I agree that the existing assert covers that. The intent was that if the log file was updated in some way that these would still be necessary. Fixed now.

tests/user/test_user_debug_logging.py
```diff
@@ -184,6 +188,98 @@ def test_add_user_request_debug_log_passphrases_and_passwords_get_redacted_on_er
...
+                    traits=TestUserConstants.TEST_ADD_USER_REQUEST_TRAITS_PASSWORD_SIMPLE,
+                )
+            except SecurityRequestError:
+                pass
+        error_log = self.ansi_escape.sub("", stdout.getvalue())
+        self.assertEqual(error_log, TestUserConstants.TEST_ADD_USER_PASSWORD_ERROR_LOG)
+        self.assertNotIn("(" + self.simple_password + ")", error_log)
+        self.assertNotIn('"' + self.simple_password + '"', error_log)
+        self.assertNotIn(">" + self.simple_password + "<", error_log)
+        self.assertIn(self.simple_password, error_log)
```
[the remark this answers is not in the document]
ElijahSwiftIBM: Fair enough, I agree that the existing assert covers that. The intent was that if the log file was updated in some way that these would still be necessary. Fixed now.

pyracf/common/security_admin.py
```diff
@@ -108,61 +132,39 @@ def _extract_and_check_result(
...
         if self.__debug:
-            self.__logger.log_xml(
-                "Result XML", result_xml, redact_strings=redact_strings
-            )
-        results = SecurityResult(
-            self.__logger.redact_strings(result_xml, redact_strings=redact_strings)
-        )
+            self.__logger.log_xml("Result XML", result_xml)
+        results = SecurityResult(result_xml)
+        del result_xml
```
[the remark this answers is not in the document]
ElijahSwiftIBM: In theory you are correct, but I also felt there was no reason to keep it either.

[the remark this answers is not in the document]
ElijahSwiftIBM: Removed per your suggestion.

pyracf/common/security_admin.py
ElijahSwiftIBM: good idea, this is cleaner. Implemented as suggested.

pyracf/common/security_admin.py
ElijahSwiftIBM: fixed

pyracf/common/security_admin.py
ElijahSwiftIBM: fixed

pyracf/common/logger.py
lcarcaramo: I am seeing the changes now.

tests/user/test_user_debug_logging.py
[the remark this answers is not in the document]
lcarcaramo: The other problem too is that in order for every possibility to be tested, you would also need to test for lowercase too. So for this test case, I think the first assertion is good enough to just makes sure that only the secret gets redacted and nothing that happens to contain the same text as the secret gets redacted too.

pyracf/common/security_admin.py
```diff
@@ -16,36 +16,9 @@ class SecurityAdmin:
...
-        "base:fvolume": "racf:fvolume",
-        "base:gaudaltr": "racf:gaudaltr",
-        "base:gaudcntl": "racf:gaudcntl",
-        "base:gaudnone": "racf:gaudnone",
-        "base:gaudread": "racf:gaudread",
-        "base:gaudupdt": "racf:gaudupdt",
-        "base:generic": "racf:generic",
-    }
+    _common_base_traits_data_set_generic = {}
+    __secret_traits = {}
```
lcarcaramo, line 20: I don't think these need to be initialized here.

pyracf/common/security_admin.py
```diff
@@ -108,61 +132,40 @@ def _extract_and_check_result(
...
             )
             self.__logger.log_xml(
                 "Request XML",
                 security_request.dump_request_xml(encoding="utf-8"),
-                redact_strings=redact_strings,
+                secret_traits=self.__secret_traits,
             )
         if self.__generate_requests_only:
-            return self.__logger.redact_strings(
+            request = self.__logger.redact_request_xml(
```
lcarcaramo, line 153: Change `request` to `request_xml`.

pyracf/common/security_admin.py
```diff
@@ -108,61 +132,40 @@ def _extract_and_check_result(
...
         )
+        self.__clear_state(security_request)
         if self.__debug:
-            self.__logger.log_xml(
-                "Result XML", result_xml, redact_strings=redact_strings
-            )
-        results = SecurityResult(
-            self.__logger.redact_strings(result_xml, redact_strings=redact_strings)
-        )
+            self.__logger.log_xml("Result XML", result_xml)
```
lcarcaramo, line 167: Might be good to have a comment here that briefly explains why this doesn't need to be redacted.

pyracf/common/logger.py
```diff
@@ -48,24 +49,25 @@ def log_dictionary(
...
         header_message: str,
         xml_string: Union[str, bytes],
-        redact_strings: List[str] = [],
+        secret_traits: dict = {},
     ) -> None:
         """Indent and colorize XML string and log it to the console."""
         if isinstance(xml_string, bytes):
             xml_string = xml_string.decode(encoding="utf-8")
-        xml_string = self.redact_strings(xml_string, redact_strings)
+        xml_string = self.redact_request_xml(xml_string, secret_traits)
```
lcarcaramo, line 70: I think we can do something like this similar to `log_dictionary` here.

```python
        if secret_traits:
            xml_string = self.redact_request_xml(xml_string, secret_traits)
```

tests/user/test_user_result_parser.py
```diff
@@ -229,8 +238,53 @@ def test_user_admin_passphrase_and_password_redacted_add_user_success_xml(
...
+            traits=TestUserConstants.TEST_ADD_USER_REQUEST_TRAITS_PASSWORD_SIMPLE,
+        )
+        self.assertEqual(
+            result,
+            TestUserConstants.TEST_ADD_USER_PASSWORD_RESULT_SUCCESS_DICTIONARY,
+        )
+        result_str = str(result)
+        self.assertNotIn("(" + self.simple_password + ")", result_str)
+        self.assertNotIn("(" + " " * len(self.simple_password) + ")", result_str)
+        self.assertIn(self.simple_password, result_str)
```
lcarcaramo, line 265: Again, I think that checking that the result dictionary you get is the same as the sample is enough here.

tests/user/test_user_result_parser.py
```diff
@@ -229,8 +238,53 @@ def test_user_admin_passphrase_and_password_redacted_add_user_success_xml(
...
+                traits=TestUserConstants.TEST_ADD_USER_REQUEST_TRAITS_PASSWORD_SIMPLE,
+            )
+        self.assertEqual(
+            exception.exception.result,
+            TestUserConstants.TEST_ADD_USER_PASSWORD_RESULT_ERROR_DICTIONARY,
+        )
+        result_str = str(exception.exception.result)
+        self.assertNotIn("(" + self.simple_password + ")", result_str)
+        self.assertNotIn("(" + " " * len(self.simple_password) + ")", result_str)
+        self.assertIn(self.simple_password, result_str)
```
lcarcaramo, line 287: Again, I think that checking that the result dictionary you get is the same as the sample is enough here.

pyracf/common/security_admin.py
lcarcaramo: I would try getting rid of these and see what happens.

tests/user/test_user_result_parser.py
ElijahSwiftIBM: This is a fair point. There are also a lot of these for regular secrets redaction and I removed these as well. They are extraneous.

tests/user/test_user_result_parser.py
ElijahSwiftIBM: This is a fair point. There are also a lot of these for regular secrets redaction and I removed these as well. They are extraneous.

pyracf/common/logger.py
ElijahSwiftIBM: Agreed. Done.

pyracf/common/security_admin.py
ElijahSwiftIBM: sure, added the following:

```python
            # No need to redact anything here since the raw result dictionary
            # already has secrets redacted when passed to logger
            self.__logger.log_xml("Result XML", result_xml)
```

pyracf/common/security_admin.py
ElijahSwiftIBM: Fair concern. Addressed.

pyracf/common/security_admin.py
ElijahSwiftIBM: Nothing broke, I'll keep them out.

pyracf/common/security_admin.py
```diff
@@ -84,6 +91,20 @@ def __overwrite_field_data(self, field_data: dict):
         """Overwrite field data in valid segment traits dictionary"""
```
lcarcaramo, line 91:
```python
        """Overwrite field data in valid segment traits dictionary."""
```

pyracf/common/security_admin.py
```diff
@@ -84,6 +91,20 @@ def __overwrite_field_data(self, field_data: dict):
...
         self._valid_segment_traits = field_data
 
+    def __add_additional_secret_traits(self, additional_secret_traits: list):
+        """Add additional fields to be redacted in logger output"""
```
lcarcaramo, line 95:
```python
        """Add additional fields to be redacted in logger output."""
```

tests/user/test_user_debug_logging.py
```diff
@@ -41,7 +42,6 @@ def test_add_user_request_debug_log_works_on_success(
             )
         success_log = self.ansi_escape.sub("", stdout.getvalue())
         self.assertEqual(success_log, TestUserConstants.TEST_ADD_USER_SUCCESS_LOG)
-        self.assertNotIn(self.test_password, success_log)
```
lcarcaramo, line 44: I think it is good to have the `self.assertNotIn()` assertions. My concern was that with those "PASSWORD_SIMPLE" tests is that the assertions on those tests added a lot of noise and in my opinion. A good assertion for those tests could be to maybe assert that you have the expected number of occurrences of `********` in debug logs and result dictionaries and that the password exists in the output to verify that only the sections that contain the actual password were redacted and that XML tags, messages, and dictionary keys that happen to contain the same string are not.

pyracf/common/security_admin.py
ElijahSwiftIBM: agree, that's a less confusing way to explain this. change made

pyracf/common/security_admin.py
ElijahSwiftIBM: That was a good catch and a broader issue. Caught and fixed in many other locations as well.

pyracf/common/security_admin.py
ElijahSwiftIBM: That was a good catch and a broader issue. Caught and fixed in many other locations as well.

tests/user/test_user_debug_logging.py
ElijahSwiftIBM: I brought back many of the previous assertions with the exception of making the described changes to debug logging. I think that counting the "********" occurrences is a novel way of tracking the "simple" password redaction in that case. I do think, though, that there isn't really a clearer way to do this in result output. For one, SMO settings will already "redact" the password, so you have to ensure that neither the unredacted password or "redacted" password appear in the output, then also ensure that a "simple" password appears elsewhere in the output. The other hang-up is that all of these items only ever appear in the "message" portion of the result dictionary anyway, so looking in specific sections is akin to just checking if an entire string matches which the first assertion really is already doing, but I like the redundancy of specific checks. I am open to discussing this further though.

pyracf/common/logger.py
```diff
@@ -93,33 +96,78 @@ def log_debug(self, header_message: str, message: str) -> None:
...
+        for trait in secret_traits:
+            segment = trait.split(":")[0]
+            if not dictionary.get(segment, {}).get(trait, {}).get("value", None):
+                continue
+            dictionary[segment][trait]["value"] = "********"
+        return dictionary
+
+    def __redact_string(
+        self,
+        input_string: Union[str, bytes],
```
lcarcaramo, line 117: I think this should just be `input_string: str` because `__redact_string` never gets a `bytes` object from any of its callers and I actually am not sure if this would work correctly with bytes objects.

tests/user/test_user_debug_logging.py
```diff
@@ -184,6 +187,88 @@ def test_add_user_request_debug_log_passphrases_and_passwords_get_redacted_on_er
...
+        with contextlib.redirect_stdout(stdout):
+            user_admin.add(
+                "squidwrd",
+                traits=TestUserConstants.TEST_ADD_USER_REQUEST_TRAITS,
+            )
+        success_log = self.ansi_escape.sub("", stdout.getvalue())
+        self.assertEqual(
+            success_log,
+            TestUserConstants.TEST_ADD_USER_ADDITIONAL_SECRET_ADDED_SUCCESS_LOG,
+        )
```
lcarcaramo, line 250: I think it probably makes sense to do an `assertNotIn` here for the additional secrets.

tests/user/test_user_debug_logging.py
```diff
@@ -184,6 +187,88 @@ def test_add_user_request_debug_log_passphrases_and_passwords_get_redacted_on_er
...
+                user_admin.add(
+                    "squidwrd",
+                    traits=TestUserConstants.TEST_ADD_USER_REQUEST_TRAITS,
+                )
+            except SecurityRequestError:
+                pass
+        error_log = self.ansi_escape.sub("", stdout.getvalue())
+        self.assertEqual(
+            error_log, TestUserConstants.TEST_ADD_USER_ADDITIONAL_SECRET_ADDED_ERROR_LOG
+        )
```
lcarcaramo, line 270: I think it probably makes sense to do an `assertNotIn` here for the additional secrets.

tests/user/test_user_debug_logging.py
ElijahSwiftIBM: Added the following for these tests:

```python
self.assertNotIn(TestUserConstants.TEST_ADD_USER_REQUEST_TRAITS["omvs:uid"], success_log)
```

tests/user/test_user_debug_logging.py
ElijahSwiftIBM: added the following for these tests:

```python
self.assertNotIn(TestUserConstants.TEST_ADD_USER_REQUEST_TRAITS["omvs:uid"], error_log)
```

lcarcaramo (approved):

Issue #21, opened 2023-10-11
Build Infrastructure: Poetry cannot install as a result of msgpack dependency being unable to build


lcarcaramo:
**Describe the bug**
`msgpack` which is part of the dependency chain for `poetry` can not install on internal build system as a result of `xlclang` not being configured correctly.

**To Reproduce**

```python
python3 -m pip install msgpack==1.0.7
```

**Expected behavior**
A description of what you expected to happen.

Latest version of `msgpack` should install with no issues.

lcarcaramo:
Resolved by moving to a different build system where clang is setup properly.

Issue #20, opened 2023-10-11
Bug Report: Secrets redaction code causes result XML that contains secrets to be truncated


lcarcaramo:
**Describe the bug**
Secrets redaction logic causes result XML to be truncated. This bug affects pyRACF version 1.0a2.

```console

                                 [pyRACF:Debug]
                                   Result XML
                            UserAdmin.set_password()


<?xml version="1.0" encoding="IBM-1047"?>
<securityresult xmlns="http://www.ibm.com/systems/zos/saf/IRRSMO00Result1">
  <user name="SQUIDWRD" operation="set" requestid="UserRequest">
    <info>Definition exists. Add command skipped due  to precheck option</info>
    <command>
      <safreturncode>0</safreturncode>
      <returncode>0</returncode>
      <reasoncode>0</reasoncode>
      <image>

Traceback (most recent call last):
  File "/usr/lpp/IBM/cyp/v3r11/pyz/lib/python3.11/xml/etree/ElementTree.py", line 1716, in close
    self.parser.Parse(b"", True) # end of data
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
xml.parsers.expat.ExpatError: no element found: line 1, column 394

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/u/<userid>/.local/lib/python3.11/site-packages/pyracf/user/user_admin.py", line 288, in set_password
    result = self.alter(userid, traits={"base:password": password})
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/u/<userid>/.local/lib/python3.11/site-packages/pyracf/user/user_admin.py", line 410, in alter
    return self._make_request(user_request, irrsmo00_precheck=True)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/u/<userid>/.local/lib/python3.11/site-packages/pyracf/common/security_admin.py", line 174, in _make_request
    results = SecurityResult(result_xml)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/u/<userid>/.local/lib/python3.11/site-packages/pyracf/common/security_result.py", line 12, in __init__
    self.__result = XMLParser.fromstring(result_xml)
[...]

Issue #19, opened 2023-09-25
Enhancement: Use poetry build backend instead of setuptools


lcarcaramo:
**Is your feature request related to a problem? Please describe.**
Setuptools does not give us the ability to publish packages to a pypi repository. Twine could be used to fill the gap, but it no longer works on z/OS because one of it's dependencies has a dependency on Rust, which is unsupported on z/OS.

**Describe the solution you'd like**
Switch to the poetry build backend.
* Use `poetry install --no-root` to install runtime/development dependencies.
* Use `poetry build` to build wheel and source tar.
* Use `poetry publish` to upload wheels to a pypi repository.
* IRRSMO00 Python extension can be built using poetry.
* After installing development dependencies with poetry, they can all be invoked using `poetry run <tool> <tool-args>`
* Poetry provides a built-in virtual environment that enables environments for multiple Python versions to be built and used.

**Describe alternatives you've considered**
Twine did work for us until one of it's dependencies was updated to use Rust, which is not supported on z/OS.

**Additional context**
The Poetry build backend should resolve all of our current build problems with the exception of pandas not working on z/OS, which will need to be addressed separately in order to integrate the reporting functionality from https://github.com/wizardofzos/pyracf

Note that pip doesn't have any support for configuring the repository source for packages and that the standard is for the user to specify the repository to use. This is almost certainly enforced for good reason, since it is probably bad security practice to allow the package metadata to specify the repository source, but it leaves pyRACF with a dependency resolution problem. Some packages will be able to be installed directly from pypi and some will need to be installed from somewhere else *(i.e., Python AI Toolkit for z/OS)*, which makes the installation procedure for pyRACF unnecessarily difficult. Regardless, a separate discussion will need to be had to discuss how to approach this particular problem.

lcarcaramo:
Resolved by PR #23

PR #23, opened 2023-10-24
Feat/poetry backend - User Administration Beta Prep - Bug fixes - Other Enhancements

lcarcaramo:
### :bulb: Issue Reference

Use Poetry Build Backend Instead of Setup tools: https://github.com/ambitus/pyracf/issues/19
Fix Secrets Redaction Bug: https://github.com/ambitus/pyracf/issues/20

This PR includes numerous updates that at a high level gets User Administration fully ready for the Beta release, converts our build backend to Poetry, fixes some bugs, and makes a few enhancements. I will do my best to cover everything here.

* GitHub actions configuration is now updated to drive everything through Poetry.
* Pre-commit hooks are now updated to use Poetry to drive everything (This may make it easier for some people to use the commit hooks).
* Updates to Contribution Guidelines for using Poetry for dependency management, building a development environment, and setting up pre-commit hooks.
* Introduce recommended branch naming conventions in Contribution Guidelines.
* Updates to Jenkinsfile to Use Poetry to drive build, test, and deploy to PyPi. This has been tested and validated, but currently does not work due to issues with the build system. For more details, please see: https://github.com/ambitus/pyracf/issues/21
* Updates to Jenkinsfile to publish source distributions to GitHub releases in addition to wheels.
* Update version number and classifiers in build metadata in anticipation of initial beta release.
* Expend all segment trait names to be clearer and more descriptive.
* Fix bug where redacted result result XML gets truncated and added a corresponding regression test.
* Fixed bug where redaction is attempted when a delete request is made for a secret trait resulting in request XML in debugging logging being displayed incorrectly along with corresponding regression tests.
* `SecurityAdmin._get_field` to has a keyword argument that enables values pulled from the result dictionary to be auto casted to strings for when a field is expected to be a string value but profile extract auto casts the value to a non-string value. This feature is fully leveraged in `UserAdmin` for getter functions that return string values.
* `_extracted_key_value_pair_segment_traits_map` class attribute added to allow short trait names in extracted segments to be expanded to more descriptive names that correspond to the `_valid_segment_traits` mappings. This functionality is implemented in `SecurityAdmin` so it can be inherited in all other function groups and used where it makes sense. This feature has been fully leveraged in `UserAdmin` for all of the initial Beta traits where it makes sense.
* Added standardized formatting of revoke date and resume date since they appear in a different date format from the julian dates, which were the only date formats that we encountered until now. This update helps ensure that all date values use the same format in extracted user profiles.
* `_case_sensitive_extracted_values` class attribute added to ensure that case sensitive traits (which are uncommon) do not get converted to lowercase. For example, an OMVS home directory of `/u/SQUIDWRD` should not be converted to all lowercase. This feature can be inherited and used in all other function groups where it makes sense.
* Profile extract now takes a segments list instead of a segments dictionary to make interface more concise, and because it is functionally useless to be able to specify `segments={"segment": False}`. A syntax of `segments=["segment]` is better.
* Full validation and implementation of getters and setters for the following user traits:
  * Revoke Date
  * Resume Date
  * Owner
  * Name
  * OMVS Max Address Space Size
  * OMVS Max CPU Time
  * OMVS Max Files Per Process
  * OMVS Max Non-Shared Memory
  * OMVS Max File Mapping Pages
  * OMVS Max Processes
  * OMVS Max Shared Memory
  * OMVS Max Threads
  * TSO Account Number
  * TSO Logon Command
  * TSO Hold Class
  * TSO Max Region Size
  * TSO Message Class
  * TSO Logon Procedure
  * TSO Default Region Size
  * TSO Sysout Class
  * TSO User Data
  * TSO Data Set Allocation Unit
* Extracted values of `"unknown"` get converted to `None`.

I think that is it.

Running the unit tests should show that everything works properly. As noted for the Poetry conversion, this has been validated, but the Jenkins build currently does not work because of a system configuration issues as noted above.

Please let me know if you have any question or if there are any issues.


ElijahSwiftIBM (approved):
All good here. We've already discussed the name changes. Beyond that the bulk of the work is Poetry setup and new gettrs and settrs.

Issue #10, opened 2023-08-11
Feature Request: Simplify Creating Resource Classes


ElijahSwiftIBM:
**Is your feature request related to a problem? Please describe.**
Even for developing tests in certain environments, creating resource classes and adjusting their attributes is a needlessly complex process. We have the building blocks in resource administration already, so let's make this more attainable.

ElijahSwiftIBM:
Resolved by #35

PR #29, opened 2023-11-02
Resource profile sub functions

ElijahSwiftIBM:
### :bulb: Issue Reference

**Issue:**  #10 

This adds 6 key sub functions to general resource profile administration:
- Class Administration
- Custom Field Administration
- Signed Program Administration
- Kerberos Realm Administration
- Started Task Administration
- APPC Session Administration

This also fixes key bugs uncovered in the new Add/Alter functions in general resource and dataset profiles as well as a workaround for unintended functionality of the IRRSMO00 callable service

The sub-functions were built like the getters/setters of resource administration focusing on specific classes and segments.

For this, unit tests were developed to verify correct generation of new requests and correct extraction of specified segment information from the proper profiles.

New Unit tests were also built to deal with the bug in Add and Alter operation errors.


pyracf/common/irrsmo00.c
```diff
@@ -13,6 +13,19 @@ typedef struct {
         char str[8];
 } VarStr_T;
 
+void null_byte_fix(char* str, unsigned int str_len) {
+   for (int i = 1; i < str_len; i++){
```
lcarcaramo, line 17: Are you able to add a comment explaining what this is for?

pyracf/resource/resource_admin.py
```diff
@@ -216,6 +270,215 @@ def get_my_access(self, resource: str, class_name: str) -> Union[str, bytes, Non
...
+            resource=signed_program_name, class_name="PROGRAM", traits=traits
+        )
+
+    def alter_signed_program(
+        self, signed_program_name: str, traits: dict = {}
+    ) -> Union[dict, bytes]:
+        """Alter an existing signed program profile."""
+        if "sigver:library" in traits:
+            traits["base:member"] = traits["sigver:library"]
+            del traits["sigver:library"]
```
lcarcaramo, line 423: What is this for?

Doesn't appear to be covered in the code coverage report.

pyracf/data_set/data_set_admin.py
```diff
@@ -110,15 +110,18 @@ def add(
             self._build_xml_segments(data_set_request)
             return self._make_request(data_set_request)
         try:
-            self.extract(data_set, volume=volume, generic=generic)
+            profile = self.extract(
```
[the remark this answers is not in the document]
ElijahSwiftIBM: LISTDSD and RLIST commands both have the capability to fall back to a more generic profile than the one you specified. RLIST does this by default, LISTDSD does not. From our perspective, this means that you can functionally get a value from a profile extract for a profile you were not looking for. I discovered this when trying to add a started task, and I was getting AddOperationError because the profile ** already existed so it was coming back from these profile extracts. This check validates the name is what is expected before throwing the error.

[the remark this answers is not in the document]
ElijahSwiftIBM: I also did add a unit test for this in Resource Admin to verify that Add and alter now avoid improperly basing information on the wrong profile. Feel free to remove the logic and run the tests to see the failures.

pyracf/common/irrsmo00.c
ElijahSwiftIBM: I don't see why not, I will try to do so to explain this better.

pyracf/resource/resource_admin.py
ElijahSwiftIBM: That's correct, I didn't write full unit tests for signed program administration as its experimental, but I wanted to build out the full code fully. I designed these subfunctions after their xml interface counterparts even though we are not using those interfaces directly.

If you review the interface for signed program (https://www.ibm.com/docs/en/zos/3.1.0?topic=types-signed-program). _The sigrequired, failload, and sigaudit fields correspond to the sigreqd, failload, sigaudit fields in the SIGVER segment of general resource profiles, as described in the [General resource administration](https://www.ibm.com/docs/en/SSLTBW_3.1.0/com.ibm.zos.v3r1.ichd100/resprof.htm) in [Reference documentation tables](https://www.ibm.com/docs/en/SSLTBW_3.1.0/com.ibm.zos.v3r1.ichd100/tabref.htm). The library field corresponds to the member field in the BASE segment._ I can remove this for now, but I thought that it would be better to build in the functionality now and debug/test later as we move to make the feature non-experimental.

pyracf/resource/resource_admin.py
```diff
@@ -180,6 +180,60 @@ def __init__(
                 "tme:roles": "racf:roles",
             },
         }
+        self._extracted_key_value_pair_segment_traits_map = {
+            "cdtinfo": {
+                "case": "caseAllowed",
+                "defaultrc": "defaultRacrouteReturnCode",
+                "first": "validFirstCharacters",
+                "generic": "generic",
+                "genlist": "genlist",
```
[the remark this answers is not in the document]
ElijahSwiftIBM: sure, this can be changed for generic and also genlist.

pyracf/data_set/data_set_admin.py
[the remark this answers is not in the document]
lcarcaramo: Thank you for the explanation. I understand the purpose of this change now.

pyracf/resource/resource_admin.py
lcarcaramo: Thank you for the explanation. I just wanted to understand what this is for. Just make sure that the documentation correctly makes the distinction between stable and experimental features and we should be good.

pyracf/common/irrsmo00.c
lcarcaramo: Also, I suggest changing the word `whitespace` to `space characters` or something like that instead if this is the case.

tests/resource/test_resource_sub_function_requests.py
```diff
@@ -0,0 +1,163 @@
...
+
+import tests.resource.test_resource_constants as TestResourceConstants
+from pyracf import ResourceAdmin
+from pyracf.common.irrsmo00 import IRRSMO00
+
+# Resolves F401
+__init__
+
+
+class TestResourceSubFunctionRequests(unittest.TestCase):
```
lcarcaramo, line 16: "Subfunction" should be one word. Changes file names and code to make "subfunction" one word.

tests/resource/test_resource_result_parser.py
```diff
@@ -63,6 +63,19 @@ def test_resource_admin_throws_error_on_add_existing_profile(
             + f"'{profile_name}' already exists as a profile in the '{class_name}' class.",
         )
 
+    def test_resource_admin_avoids_error_on_add_covered_profile(
```
lcarcaramo, line 66: What does `covered_profile` mean?

ElijahSwiftIBM: The intent is to mean a profile "covered" (or protected) by a more generic one. In this case, TEST* already exists which would appear on RLIST TESTING as TEST* does "cover" the TESTING resource (as well as anything else that starts with TEST and has no lower qualifiers). This is a word I have thrown around on L2 support a lot, but I'm not sure if it came from documentation/standards or just my understanding of how profiles work.

tests/resource/test_resource_sub_function_requests.py
ElijahSwiftIBM: Sure; done.

pyracf/common/irrsmo00.c
ElijahSwiftIBM: I suppose we don't know what the exact EBCDIC characters are returned by the RACF command as I've only ever copy/pasted what's rendered by ISPF. I will change the wording to make it generic enough to relay that, at least until we can discuss with RACF development to see if it is a "bug" or not.

tests/resource/test_resource_result_parser.py
lcarcaramo: I see this is for testing when there is already a more generic profile protecting the resource. Should there be a similar test for data set admin or do we think this is good enough?

ElijahSwiftIBM: The same edge case exists, but if I am not mistaken it requires driving the GENERIC keyword on LISTDSD (which maps to the base:generic trait on dataset admin command) so it is not as likely to appear especially with the lack of subfunctions or settrs. I think it should probably get a test case for completeness, but I wasn't rushing to get it into the beta with the other features. I will try to add it today before this branch gets merged.

pyracf/common/irrsmo00.c
lcarcaramo:
```c
// This function changes any null character not preceded by '>' to a blank character.
// This is a workaround for an issue where the xml returned by IRROSMO00 sometimes
// includes null characters instead of properly encoded text, which causes the returned
// xml to be truncated.
```

tests/resource/test_resource_sub_function_requests.py
lcarcaramo: Please update the filenames as well.

ElijahSwiftIBM: done;

lcarcaramo (approved):

Issue #30, opened 2023-11-02
Bug Report: Extra Messages Not Removed From Profile Extract Data


lcarcaramo:
**Describe the bug**
If extra messages get added to profile extract data in result xml returned by IRRSMO00, it should not be included in the Python dictionary that the XML gets transformed into.

**To Reproduce**

* Logon to a z/OS user id that is only authorized to extract their own base segment but no other segments.
* Ensure an affected version of pyRACF is installed.
* Extract the user's own profile and include additional segments.
```python
from pyracf import UserAdmin
user_admin = UserAdmin()
# An even number of additional segments results in extraneous
# malformed data being added to the extracted profile data
user_admin.extract("squidwrd", segments=["omvs", "tso"]
# An odd number of additional segments results in a stack trace.
user_admin.extract("squidwrd", segments=["omvs"]
```

**Expected behavior**
All profile extract functions or more specifically anything that executes the `listdata` operation, should exclude any extra messages attached to the end of the result xml.

**Console Output**

Two Additional Segments:
```python
user_admin.extract("squidwrd", segments=["omvs", "tso"]
```

Result XML
```xml
<?xml version="1.0" encoding="IBM-1047"?>
<securityresult xmlns="http://www.ibm.com/systems/zos/saf/IRRSMO00Result1">
  <user name="SQUIDWRD" operation="listdata" requestid="UserRequest">
    <command>
      <safreturncode>0</safreturncode>
      <returncode>0</returncode>
      <reasoncode>0</reasoncode>
      <image>LISTUSER SQUIDWRD  OMVS     TSO      OVM     </image>
      <message>USER=SQUIDWRD  NAME=SQUIDWARD             OWNER=LEONARD   CREATED=23.094</message>
      <message> DEFAULT-GROUP=SYS1     PASSDATE=00.000 PASS-INTERVAL=186 PHRASEDATE=N/A</message>
      <message> ATTRIBUTES=NONE</message>
      <message> REVOKE DATE=NONE   RESUME DATE=NONE</message>
      <message> LAST-ACCESS=23.094/12:55:37</message>
      <message> CLASS AUTHORIZATIONS=NONE</message>
      <message> NO-INSTALLATION-DATA</message>
      <message> NO-MODEL-NAME</message>
      <message> LOGON ALLOWED   (DAYS)          (TIME)</message>
      <message> ---------------------------------------------</message>
      <message> ANYDAY                          ANYTIME</message>
[...]

PR #31, opened 2023-11-02
Bug/filter out tso messages on success

lcarcaramo:
### :bulb: Issue Reference

**Issue:**  https://github.com/ambitus/pyracf/issues/30

Resolves https://github.com/ambitus/pyracf/issues/30 and removes extraneous code from the `test_alter_user_request_debug_log_password_xml_tags_not_redacted_on_error()` test case in `test_user_debug_logging.py`.

* Extra messages in profile extract data or more specifically data returned by the `listdata` operation are removed. This is implemented in `common/security_result.py` so that this bug is resolved for all function groups that use the `listdata` operation.
* Extraneous code has been removed from `test_alter_user_request_debug_log_password_xml_tags_not_redacted_on_error()`.

See results of Unit Tests in Github Actions.

### DIFF
--- a/pyracf/common/irrsmo00.c
+++ b/pyracf/common/irrsmo00.c
@@ -14,7 +14,7 @@ typedef struct {
 } VarStr_T;
 
 // This function changes any null character not preceded by '>' to a blank character.
-// This is a workaround for an issue where the profile data embedded in response xml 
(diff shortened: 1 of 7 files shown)


ElijahSwiftIBM (approved):
Looks good!

Issue #33, opened 2023-11-02
Bug Report: SegmentError and SegmentTraitError keep built dictionaries


ElijahSwiftIBM:
**Describe the bug**
SegmentError and SegmentTraitError exit the code without clearing segment trait dictionaries. This leads to improper request dictionaries built on the next request.

**To Reproduce**
Make a request that will encounter SegmentTraitError or SegmentError. Make another request incompatible with those traits.

**Expected behavior**
These requests should be independent of one another if at all possible

**Console Output**
N/A

**Environment Information:**
 - IBM Python 3.11 on z/OS 2.4

**Additional context**
N/A

lcarcaramo:
Resolved by: #34

PR #34, opened 2023-11-02
Clear State before raising errors

ElijahSwiftIBM:
Add state clearing before raising SegmentError or SegmentTraitError.

Add unit test for this (Unit test for SegmentTraitError includes extract to simulate functionality of add when generateRequestOnly is not set)

### :bulb: Issue Reference

**Issue:**  #33 

SegmentError and SegmentTraitError exit the code without clearing segment trait dictionaries. This leads to improper request dictionaries built on the next request.

added call to clear_state() function before throwing errors.

Built new test cases, one for SegmentError and one for SegmentTraitError, both hanging off of USER Admin

Test cases work as designed. Without clear_state() both fail, with it they succeed.

### DIFF
--- a/tests/user/test_user_request_builder.py
+++ b/tests/user/test_user_request_builder.py
@@ -140,6 +140,9 @@ class TestUserRequestBuilder(unittest.TestCase):
             + f"combination for '{self.user_admin._profile_type}'.\n",
         )
 
+# Since this test uses GenerateRequestsOnly, the "Add" after the AddOperationError is returned
+# does not begin with an "Extract" call. This is necessary to recreate the error, so an extra
+# extract call was added to simulate this behavior.
     def test_user_admin_cleans_up_after_build_add_request_with_bad_segment_traits(self):
         bad_trait = "omvs:bad_trait"
         user_admin = UserAdmin(
--- a/tests/user/test_user_request_builder.py
+++ b/tests/user/test_user_request_builder.py
@@ -140,6 +140,30 @@ class TestUserRequestBuilder(unittest.TestCase):
             + f"combination for '{self.user_admin._profile_type}'.\n",
         )
 
+    def test_user_admin_cleans_up_after_build_add_request_with_bad_segment_traits(self):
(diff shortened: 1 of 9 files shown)


tests/user/test_user_request_builder.py
```diff
@@ -140,6 +140,30 @@ def test_user_admin_build_add_request_with_bad_segment_traits(self):
...
+            user_admin.add(
+                "squidwrd", TestUserConstants.TEST_ADD_USER_REQUEST_BAD_TRAITS
+            )
+        self.assertEqual(
+            exception.exception.message,
+            "Unable to build Security Request.\n\n"
+            + f"'{bad_trait}' is not a known segment-trait "
+            + f"combination for '{self.user_admin._profile_type}'.\n",
+        )
+        result = user_admin.extract("squidwrd", segments=["omvs"])
```
lcarcaramo, line 158: Is there a reason for doing extract and add here?

ElijahSwiftIBM: Yes, in GenerateRequestOnly Mode, Add does not do a preliminary extract, but outside of this mode it does to know if it needs to throw AddOperationError(s). This extract emulates that, as this is necessary to see the impact of the bug.

lcarcaramo: Can you add a comment?

tests/user/test_user_request_builder.py
```diff
@@ -140,6 +140,33 @@ def test_user_admin_build_add_request_with_bad_segment_traits(self):
             + f"combination for '{self.user_admin._profile_type}'.\n",
         )
 
+# Since this test uses GenerateRequestsOnly, the "Add" after the AddOperationError is returned
+# does not begin with an "Extract" call. This is necessary to recreate the error, so an extra
+# extract call was added to simulate this behavior.
```
lcarcaramo, line 145: Indent comment.

lcarcaramo (approved):

Issue #37, opened 2023-11-03
Bug Report: Resource Profile Subfunctions extract "" as "" and not as null


ElijahSwiftIBM:
**Describe the bug**
Resource Profile Subfunctions extract "" as "" and not as null

**To Reproduce**
Set a trait like grouping_class_name to "" in a resource profile and extract it with any applicable profile extract function.

**Expected behavior**
This should be cast to null rather than left as an empty string

**Console Output**
Visible in unit tests and documentation for resource profile subfunctions

**Environment Information:**
N/A

**Additional context**
N/A

lcarcaramo:
resolved by #38 and #39

PR #38, opened 2023-11-07
Bug/resource admin fixes

ElijahSwiftIBM:
### :bulb: Issue Reference

**Issue:**  #37 

Resource subfunctions were extracting "" instead of `None`. Technically this was not a problem as this is still a falsy value, but we wanted to make this more consistent.

Additionally, testing and bug fixes were done to stabilize the APPC Session Extract function and additional traits for resource classes.

Minor changes to Profile Extract in the security admin object drives "" to `None`. Additionally a mistake was fixed in resource admin's trait map and logic was added to resource admin's profile extract to handle the language in the APPC Session output that was not parsing properly before.

There are no new test cases, but existing ones were modified to accommodate the code adjustments ("" became `None` where possible) and the test case for APPC Session Extract was bolstered to parse real data successfully.


pyracf/common/security_admin.py
```diff
@@ -440,6 +441,10 @@ def __format_data_set_generic_profile_data(
             i += 1
         elif "NO INSTALLATION DATA" in messages[i]:
             profile[current_segment]["installationData"] = None
+        elif "UNLOCKED" in messages[i]:
+            profile[current_segment]["locked"] = False
+        elif "LOCKED" in messages[i]:
+            profile[current_segment]["locked"] = True
```
lcarcaramo, line 447: Is there a test to validate this?

lcarcaramo: Nevermind. I see it. This is for APPC session.

ElijahSwiftIBM: Yeah, the actual code for the test did not change, but the constants did since previously the only trait being tested was not properly read in.

pyracf/common/security_admin.py
```diff
@@ -683,6 +688,10 @@ def _clean_and_separate(self, value: str) -> Union[list, str]:
         if isinstance(out, list):
             if None in out:
                 return None
+            if "days" in out:
+                del out[out.index("days")]
+                if len(out) == 1:
+                    return out[0]
```
lcarcaramo, line 694: What is this change for?

ElijahSwiftIBM: In APPC Session, there's a field for the interval where it is written to output as "x days". This allowed the resulting dictionary to have an integer value instead of appearing as [x, 'days'] which is how it came through with the previous logic. We had a similar issue in SETROPTS, but we solved it differently, and the solution couldn't be applied directly here.

lcarcaramo: Can you post a before and after sample of what this looks like in profile extract?

ElijahSwiftIBM: The unit testing covers the "After" case:

```
      <message>SESSION INFORMATION</message>
      <message>-------------------</message>
      <message>SESSION KEY IS E3C5E2E3D2C5E800</message>
      <message>SESSION KEY INTERVAL IS 00005 DAYS</message>
      <message>LOCKED</message>
      <message>CONVSEC= CONV</message>
```

becomes:
```
{
    "sessionKeyInterval": 5,
    "locked": True,
    "sessionKey": "e3c5e2e3d2c5e800",
    "securityCheckingLevel": "conv",
}
```

but before this change it was:
```
  {'locked': True,
   'securityCheckingLevel': 'conv',
   'sessionKey': 'e3c5e2e3d2c5e800',
 'sessionKeyInterval': [5, 'days']}
```

ElijahSwiftIBM: The associated logic is that the " IS " functionally becomes treated like an "=" where "SESSION KEY INTERVAL" becomes the key and "00005 DAYS" becomes the value. The existing code sees the " " there and breaks the value into a list, which correctly casts "00005" to an int 5, but keeps the string "days" in the list which is no longer necessary because there was no logic here to throw it out. SETROPTS dealt with "days" in a separate function so the solution didn't transfer well.

lcarcaramo: Thank you! I will add a comment about this to the release notes.

lcarcaramo: I don't know if it makes sense to turn the key into `sessionKeyIntervalDays` or if it can be assumed that the user knows that the unit is days?

lcarcaramo: I think I had it set to `sessionKeyVerificationInterval` in `SetroptsAdmin`. I don't know if you think it makes sense to change it to that to make everything consistent? On the other hand we could also update Setropts later to be consistent with General Resource Profile Admin. I don't know what you think?

I also had `sessionKeyVerificationInterval` under a section called `vtam` for `SetroptsAdmin.list_racf_options()` and it may actually be more accurate to put it under a section called `appc`.

This is one of the reasons why I wanted to mark the `SetroptsAdmin.list_racf_options()` function as experimental even though it is more or less complete and working at this point.

ElijahSwiftIBM: Yes, I would say that whatever you want to call it, I think living under appc or session would make more sense in the setropts breakdown. I would also say that for both the trait and the extract key, there should be a "max" in there somewhere as the global setting is the maximum value the individual settings can use.

lcarcaramo: So you are thinking that it should be something like `maxSessionKeyVerificationInterval` everywhere basically?

ElijahSwiftIBM: personally, in reading both descriptions, I don't think the word verification adds much to this I would be in favor of either `sessionKeyInterval `in resource admin and `maxSessionKeyInterval `in setropts or `sessionKeyVerificationInterval `in resource and `maxSessionKeyVerificationInterval` in setropts.

lcarcaramo: So, if you don't think the word `verification` helps people better understand the purpose of the trait very much, then I am in favor of removing the word `verification`. In that case it sounds like we can leave `sessionKeyInterval` as is in General Resource Admin, but we may want to change `sessionKeyVerificationInterval` in Setropts Admin to `maxSessionKeyInterval` since it is defining the max session key interval allowed for general resource profiles protecting APPC sessions in addition to changing the `vtam` key to `appc` in `SetroptsAdmin.list_racf_options()`.

ElijahSwiftIBM: agreed, I can work those changes into this pull request and make any associated documentation changes in the active doc PR for this issue.

lcarcaramo: Sure. if you are able to, I think it would be good to make those changes now instead of adding it to the backlog.

ElijahSwiftIBM: Corresponding doc changes made as well.

lcarcaramo (approved):

Issue #40, opened 2023-11-09
Bug Report: Resource Profile Extract Auditing Issues


ElijahSwiftIBM:
**Describe the bug**
When Resource Profiles are extracted, the Auditing trait is not correctly parsed, leading to invalid data in the dictionary.

**To Reproduce**
Add or alter a resource such that it has multiple Auditing values such as: "SUCCESS(READ),FAILURES(UPDATE)" in the example, the result dictionary showed:

```
                "auditing": {
                  "success": "update"
                }
```

**Expected behavior**
Result dictionary should show:

```
                "auditing": {
                  "success": "read",
                  "failures":"update"
                }
```

**Console Output**

                                                                               
                                 [pyRACF:Debug]                                
                               Result Dictionary                               
                            ResourceAdmin.extract()                            
                                                                               

{
  "securityResult": {
    "resource": {
      "name": "TSTRES2",
      "class": "ELIJTEST",
      "operation": "listdata",
      "requestId": "ResourceRequest",
      "commands": [
        {
          "safReturnCode": 0,
          "returnCode": 0,
          "reasonCode": 0,
          "image": "RLIST   ELIJTEST             (TSTRES2) ",
          "messages": [
            "CLASS      NAME",
            "-----      ----",
            "ELIJTEST   TSTRES2",
            " ",
            "LEVEL  OWNER      UNIVERSAL ACCESS  YOUR ACCESS  WARNING",
            "-----  --------   ----------------  -----------  -------",
            " 00    ESWIFT          NONE               NONE    NO",
            " ",
            "INSTALLATION DATA",
            "-----------------",
            "NONE",
            " ",
            "APPLICATION DATA",
            "----------------",
            "NONE",
            " ",
            "AUDITING",
            "--------",
            "SUCCESS(READ),FAILURES(UPDATE)",
            " ",
            "NOTIFY",
            "------",
            "NO USER TO BE NOTIFIED"
          ]
        }
      ]
    },
    "returnCode": 0,
    "reasonCode": 0
  }
}

                                                                               
                                 [pyRACF:Debug]                                
                     Result Dictionary (Formatted Profile)                     
                            ResourceAdmin.extract()                            
                                                                               

{
  "securityResult": {
    "resource": {
      "name": "TSTRES2",
      "class": "ELIJTEST",
      "operation": "listdata",
      "requestId": "ResourceRequest",
      "commands": [
        {
          "safReturnCode": 0,
          "returnCode": 0,
[...]

PR #51, opened 2023-11-21
Bug/resource extract auditing

ElijahSwiftIBM:
### :bulb: Issue Reference

**Issue:**  #40 

Audit rules were not properly handled by General Resource profile extract. There can be multiple entries in this category.

Additionally, new features were added to manipulate these audit rules to provide smoother test interfaces and future use of these traits.

Logic was changed in general resource profile extract to break apart multi-entry sections that present like Auditing Rules.

Added several getters and setters including new compound setters that use information obtained in an extract to apply proper traits in an alter.

I edited existing test cases for profile extract to use an audit rules value that would have failed in the previous implementation.

New test cases were added for the getters and setters including a new test case type that leverages debug logging as the compound setters needed to verify both request building and result parsing while mocking in the initial extract request, so the conventional architecture of generaterequestonly could not fully satisfy this.


pyracf/common/security_admin.py
```diff
@@ -222,6 +222,10 @@ def _to_steps(self, results: Union[List[dict], dict, bytes]) -> Union[dict, byte
         steps_dictionary = {}
         for step, result_dictionary in enumerate(pre_processed_results):
             steps_dictionary[f"step{step+1}"] = result_dictionary
+        self.__logger.log_dictionary(
+            "Steps Dictionary",
+            steps_dictionary,
+        )
```
lcarcaramo, line 228: I don't know if displaying the security result steps dictionary is necessary. All security result dictionary are already displayed in the logs before this. I am not sure if this adds much and if we do keep it, there needs to be a conditional that checks if `self.__debug` is `True` before attempting to log the dictionary.

pyracf/resource/resource_admin.py
```diff
@@ -265,11 +265,316 @@ def set_universal_access(
...
+        return self._get_field(profile, "base", "auditing")
+
+    def clear_audit_failures(
+        self, resource: str, class_name: str
+    ) -> Union[dict, bytes]:
+        """
+        Clears the auditing rules for failures for this general resource profile while
+        preserving other rules.
+        """
+        result = [self.extract(resource, class_name)]
```
lcarcaramo, line 293: Profile extract results should not be in the security result steps dictionary. The idea I had with the security result steps dictionary was that for any compound function, it would serve as an audit trail of all operations that make alterations. Since profile extract just retrieves information and doesn't modify anything, it should be left out.

pyracf/resource/resource_admin.py
```diff
@@ -265,11 +265,316 @@ def set_universal_access(
...
+        profile = result[0]["securityResult"]["resource"]["commands"][0]["profiles"][0]
+        audit_rules = self._get_field(profile, "base", "auditing")
+        if "failures" not in audit_rules:
+            return self._to_steps(result[0])
+        traits = {}
+        if "success" in audit_rules:
+            traits[f"base:audit_{audit_rules['success']}"] = "SUCCESS"
+        if "all" in audit_rules:
+            traits[f"base:audit_{audit_rules['all']}"] = "ALL"
```
lcarcaramo, line 302: If there is currently an audit rule for failures shouldn't there be a trait that unsets that, or is it that anytime that you set any audit rules, all existing audit rules will be overwritten with what you specify?

pyracf/resource/resource_admin.py
```diff
@@ -265,11 +265,316 @@ def set_universal_access(
...
+            return self._to_steps(result[0])
+        traits = {}
+        if "failures" in audit_rules:
+            traits[f"base:audit_{audit_rules['failures']}"] = "FAILURE"
+        if "all" in audit_rules:
+            traits[f"base:audit_{audit_rules['all']}"] = "ALL"
+        result.append(self.alter(resource, class_name, traits=traits))
+        return self._to_steps(result)
+
+    def clear_audit_both_successes_and_failures(
```
lcarcaramo, line 326: I think you can remove `both`

```python
def clear_audit_successes_and_failures(
```

pyracf/common/security_admin.py
ElijahSwiftIBM: That is a totally fair point, I can scrap this. I was thinking we would need it for unit testing, but you are right that all the steps were already displayed in the log.

pyracf/resource/resource_admin.py
ElijahSwiftIBM: Ah, see I see how that design intent would be valuable, but I could also see how the steps dictionary could have value as an educational tool to understand the operations pyRACF is doing under the covers. I agree more with your stated intent, though and I will change this now.

pyracf/resource/resource_admin.py
ElijahSwiftIBM: I think this is covered by our discussion of the functionality and re-design of these functions. Let me know if this is still a concern.

pyracf/resource/resource_admin.py
ElijahSwiftIBM: I think this is covered by our discussion of the functionality and re-design of these functions. Let me know if this is still a concern.

pyracf/resource/resource_admin.py
```diff
@@ -285,6 +285,134 @@ def get_user_access(
...
+        if control is not None:
+            traits["base:audit_control"] = control
+        if read is not None:
+            traits["base:audit_read"] = read
+        if update is not None:
+            traits["base:audit_update"] = update
+        result = self.alter(resource, class_name, traits=traits)
+        return self._to_steps(result)
+
+    def clear_all_audit_rules(
```
lcarcaramo, line 409: I think it might make sense to rename `clear_all_audit_rules` to `remove_all_audit_rules` because it is more consistent with other function names like `remove_all_class_authorizations`.

tests/resource/test_resource_getters.py
```diff
@@ -60,6 +60,9 @@ def test_resource_admin_get_universal_access_raises_an_exception_when_extract_fa
         with self.assertRaises(SecurityRequestError):
             self.resource_admin.get_universal_access("TESTING", "ELIJTEST")
 
+    # ============================================================================
+    # My Access
+    # ============================================================================
     def test_resource_admin_get_my_access_returns_valid_when_read(
```
lcarcaramo, line 66: I think the following naming convention is cleaner here:

```python
def test_resource_admin_get_my_access_read(self):
def test_resource_admin_get_my_access_none(self):
def test_resource_admin_get_my_access_raises_an_exception_when_extract_fails(self):
def test_resource_admin_get_audit_rules(self):
def test_resource_admin_get_audit_rules_raises_an_exception_when_extract_fails(self):
```

pyracf/resource/resource_admin.py
ElijahSwiftIBM: agree this is better. change in next commit.

tests/resource/test_resource_getters.py
ElijahSwiftIBM: agree this is better. change in next commit.

pyracf/resource/resource_admin.py
lcarcaramo: For the educational aspect of things here, I think the debug logging fills that role.

pyracf/resource/resource_admin.py
lcarcaramo: Ok, so based on our conversation on this functionality, my understanding of how overwrite works is that we don't need to explicitly clear any audit rules because the act of creating new audit rules deletes any audit rules that exist, which makes this question irrelevant.

pyracf/resource/resource_admin.py
lcarcaramo: Correct. We no longer have a clear function because the the setters can be used to do this.

tests/resource/resource_log_samples/alter_resource_alter_audit_by_audit_alter_access_error.log
```diff
@@ -0,0 +1,70 @@
+
```
lcarcaramo, line 1: All of these debug log files should be able to be deleted now that we are using mock to verify the behavior of the audit rules setters using mocks and the generated security request dictionaries, no?

pyracf/resource/resource_admin.py
```diff
@@ -285,6 +285,133 @@ def get_user_access(
...
+        return self._get_field(profile, "base", "auditing")
+
+    def overwrite_audit_by_attempt(
+        self,
+        resource: str,
+        class_name: str,
+        success: Union[str, None] = None,
+        failure: Union[str, None] = None,
+        all: Union[str, None] = None,
+    ) -> Union[dict, bytes]:
```
lcarcaramo, line 305: I don't know if it makes sense to verify that the values the user gives here actually results in valid keys here? If someone specificies `success="doodlebob`, then we will end up with the trait `"base:audit_doodlebob" = "success`. The error will be caught downstream, but we it probably makes sense to check this early and say something like this:

```python
success = success.lower()
if success not in ("none", "read", "alter", "control"):
    # raise a new type of exception?
```

lcarcaramo: For access level, I think we could probably let the downstream error handling handle it, but for attempt, I think it could make sense to do a check her since the value that the user gives ends up becoming a trait key. Downstream error handling should be executed to pick that up, but the error might be more confusing for the user in that case.

tests/resource/resource_log_samples/alter_resource_alter_audit_by_audit_alter_access_error.log
ElijahSwiftIBM: yes, sorry. I forgot to take them out after refactoring.

pyracf/resource/resource_admin.py
ElijahSwiftIBM: That makes a lot of sense. I added functionality to raise a traditional `ValueError` and added unit testing that ensures this is raised for these functions.

pyracf/resource/resource_admin.py
```diff
@@ -589,3 +718,43 @@ def _format_profile(self, result: dict) -> None:
...
+    ):
+        access_levels = ["alter", "control", "read", "update"]
+        value_error_text = (
+            "Please specify 'alter', 'control', 'read', or 'update' access "
+            + "for your target attempt."
+        )
+        bad_values = []
+        if success is not None:
+            if success.lower() not in access_levels:
+                bad_values.append(success)
```
lcarcaramo, line 736: I think you can do this to consolidate things a little bit:

```python
    if (
            success is not None
            and not isinstance(success, str)
            and success.lower() not in access_levels
    ):
        bad_values.append(str(success))
```

Also it is probably a good idea to make sure the value is a string too.

pyracf/resource/resource_admin.py
```diff
@@ -589,3 +718,43 @@ def _format_profile(self, result: dict) -> None:
...
+            if failure.lower() not in access_levels:
+                bad_values.append(failure)
+        if all is not None:
+            if all.lower() not in access_levels:
+                bad_values.append(all)
+        if len(bad_values) == 0:
+            return
+        elif len(bad_values) == 1:
+            value_error_text = (
+                f"'{bad_values[0]}' is not a proper value. " + value_error_text
```
lcarcaramo, line 747:
```python
f"'{bad_values[0]}' is not a valid access level. " + value_error_text
```

Similar updates should be made to the other error text conditions as well.

pyracf/resource/resource_admin.py
```diff
@@ -589,3 +718,43 @@ def _format_profile(self, result: dict) -> None:
 
         del result["securityResult"]["resource"]["commands"][0]["messages"]
         result["securityResult"]["resource"]["commands"][0]["profiles"] = profiles
+
+    def __validate_access_level(
```
lcarcaramo, line 722: Should this be called `__valid_access_level()` since we are potentially validating several access levels.

tests/resource/test_resource_getters.py
```diff
@@ -95,3 +98,29 @@ def test_resource_admin_get_my_access_raises_an_exception_when_extract_fails(
         )
         with self.assertRaises(SecurityRequestError):
             self.resource_admin.get_my_access("TESTING", "ELIJTEST")
+
+    # ============================================================================
+    # Auditing Rules
+    # ============================================================================
+    def test_resource_admin_get_audit_rules(
```
lcarcaramo, line 105: Does it make sense to have a tests for extracting all possible values?

lcarcaramo: Or is this all possible values?

lcarcaramo: I don't know if there is an `all` key that we could end up with here.

tests/resource/test_resource_setters.py
```diff
@@ -18,10 +18,258 @@ class TestResourceSetters(unittest.TestCase):
...
+            TestResourceConstants.TEST_EXTRACT_RESOURCE_GET_AUDIT_RULES
+        )
+        self.assertEqual(
+            self.resource_admin.alter_audit_by_attempt(
+                "TESTING", "ELIJTEST", failure="control", success="alter", all="read"
+            ),
+            TestResourceConstants.TEST_RESOURCE_ALTER_AUDIT_BY_ATTEMPT_ALL_REQUEST_XML,
+        )
+
+    def test_resource_admin_build_alter_audit_by_attempt_value_error(self):
```
lcarcaramo, line 256: We should probably keep the `ValueError` tests together, no?

tests/resource/test_resource_setters.py
```diff
@@ -18,10 +18,258 @@ class TestResourceSetters(unittest.TestCase):
...
+        bad_success = "problem"
+        bad_failure = "improper"
+        bad_all = "bad"
```
lcarcaramo, line 259: It might be good to set one of these values to an integer or something to make sure your checks can handle that. For example, running the `.lower()` on an integer will result in an `AttributeError` and you want to make sure your code can handle that.

pyracf/resource/resource_admin.py
```diff
@@ -589,3 +718,43 @@ def _format_profile(self, result: dict) -> None:
...
+        self,
+        success: Union[str, None] = None,
+        failure: Union[str, None] = None,
+        all: Union[str, None] = None,
+    ):
+        access_levels = ["alter", "control", "read", "update"]
```
lcarcaramo, line 728: Make this a tuple:

```python
access_levels = ("alter", "control", "read", "update")
```

pyracf/resource/resource_admin.py
lcarcaramo: You may even be able to turn this into a for each loop as well:

```python
    for access_level in (success, failure, all):
        if (
                access_level is not None
                and not isinstance(access_level, str)
                and access_level.lower() not in access_levels
        ):
            bad_values.append(str(access_level))
```

pyracf/resource/resource_admin.py
lcarcaramo: May also make sense to rename this to `valid_access_levels`.

pyracf/resource/resource_admin.py
```diff
@@ -589,3 +718,43 @@ def _format_profile(self, result: dict) -> None:
...
+        value_error_text = (
+            "Please specify 'alter', 'control', 'read', or 'update' access "
+            + "for your target attempt."
+        )
+        bad_values = []
```
lcarcaramo, line 733: Change name to `bad_access_levels`?

pyracf/common/security_admin.py
```diff
@@ -668,6 +650,44 @@ def __format_semi_tabular_data(
                 messages[i + 2][indexes[j] : ind_e1]
             )
 
+    def __format_tabular_data(
```
lcarcaramo, line 653: Does this code fix a bug or is this just refactoring?

lcarcaramo: Based on the description for the PR this fixes a problem with profile extract, right?

pyracf/resource/resource_admin.py
```diff
@@ -589,3 +718,43 @@ def _format_profile(self, result: dict) -> None:
...
+        self,
+        success: Union[str, None] = None,
+        failure: Union[str, None] = None,
+        all: Union[str, None] = None,
+    ):
+        access_levels = ["alter", "control", "read", "update"]
+        value_error_text = (
+            "Please specify 'alter', 'control', 'read', or 'update' access "
+            + "for your target attempt."
+        )
```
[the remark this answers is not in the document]
ElijahSwiftIBM: sure. change made

pyracf/resource/resource_admin.py
ElijahSwiftIBM: That was a much better and cleaner idea. I did not include checking to see if it is a string because anything that is not a string will be cast into one by the format strings or the new code as follows:

```python3
        for attempt_argument in (success, failure, all):
            if (
                attempt_argument is not None
                and str(attempt_argument).lower() not in access_levels
            ):
                bad_values.append(attempt_argument)
```

pyracf/resource/resource_admin.py
ElijahSwiftIBM: I wanted to stay away from calling these access levels in messaging even though they are, because this is for the `by_attempt` functions. I think it will be confusing to have people see those terms intermingling even though its true. The `by_attempt` functions you use the attempt as the argument and the value you pass in is an access level and for the `by_access_level` functions the access level is the argument and you pass in an attempt type.

pyracf/resource/resource_admin.py
ElijahSwiftIBM: sure, change made.

tests/resource/test_resource_getters.py
ElijahSwiftIBM: I can add more tests that pull out other combinations of values. All can also appear. Due to the nature of the interactions of the rules I have not yet generated a ruleset with ALL SUCCESS and FAILURE all at the same time, but I will try to do so to make this more comprehensive.

tests/resource/test_resource_setters.py
ElijahSwiftIBM: I viewed this less as a testing of the value error functionality and more as a test of the `alter_audit_by_attempt` function to ensure it produces a value error under the right conditions. Functionally it tests both, so i think a good case can be made for either. (I did one for alter and one for overwrite).

tests/resource/test_resource_setters.py
ElijahSwiftIBM: new code handles that and I changed the test to take in an int for `bad_all` to verify that is handled as expected.

pyracf/resource/resource_admin.py
ElijahSwiftIBM: sure, changes made.

pyracf/resource/resource_admin.py
ElijahSwiftIBM: sure. change made.

pyracf/resource/resource_admin.py
```diff
@@ -589,3 +718,43 @@ def _format_profile(self, result: dict) -> None:
...
+        if success is not None:
+            if success.lower() not in access_levels:
+                bad_values.append(success)
+        if failure is not None:
+            if failure.lower() not in access_levels:
+                bad_values.append(failure)
+        if all is not None:
+            if all.lower() not in access_levels:
+                bad_values.append(all)
+        if len(bad_values) == 0:
```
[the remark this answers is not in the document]
ElijahSwiftIBM: I didn't realize they finally added switch statements to python. I will make this change now.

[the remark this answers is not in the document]
ElijahSwiftIBM: I also refactored code for the "default" case to make it more generic, in case this ever needs to be expanded or generalized to handle more arguments.

pyracf/common/security_admin.py
ElijahSwiftIBM: Yes. This change fixed the bug described in the initial issue for this PR and then while I was playing with those settings I added features. The "tabular data" allows the code to grab multiple key-value pairs on one like of text when they use commas to split multiple items under one "header" like with AUDITING.

lcarcaramo: And given that this function is specific to general resource profile, it does not potentially affect anything else.

lcarcaramo: Thank you for the confirmation!

pyracf/resource/resource_admin.py
lcarcaramo: I still think it is still probably a good idea to to check if the value is a string just to ensure that nothing strange happens if it gets to `and str(attempt_argument).lower() not in access_levels`.

pyracf/resource/resource_admin.py
lcarcaramo: I disagree. The `by_atempt` functions take three key-value pairs in the format `attempt=access_level`. Saying that the argument provided `is not a valid access level.` is more specific than `is not a proper value`. It also has the added benefit of being an educational aid since it teaches the user that the things that they are specifying as parameters are access levels.

tests/resource/test_resource_getters.py
lcarcaramo: So, if there are a few other permutations of audit rules that have the potential to break things, I definitely think it is worth looking into adding a couple more tests for those.

tests/resource/test_resource_setters.py
lcarcaramo: I think that is fine. Only other suggestion that I would have is to maybe create a test the second case on the switch statement in `__validate_access_levels`.

pyracf/resource/resource_admin.py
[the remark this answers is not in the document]
lcarcaramo: For the default case, I think you can do this instead to clean things up:

```python
           case _:
                bad_access_levels = [
                    f"'{bad_access_level}'" for bad_access_level in bad_access_levels
                ]
                bad_access_levels[-1] = f"and {bad_access_levels[-1]} "
                value_error_text = (
                    f"{', '.join(bad_access_levels)} are not proper values. {value_error_text}"
                )
```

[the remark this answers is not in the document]
lcarcaramo: And for `case 2`, you can do:

```python
            case 2:
                value_error_text = (
                    f"'{bad_access_levels[0]}' and '{bad_access_levels[1]}' are not proper "
                    + f"values. {value_error_text}"
                )
```

pyracf/resource/resource_admin.py
ElijahSwiftIBM: sounds good, you have convinced me.

tests/resource/test_resource_setters.py
ElijahSwiftIBM: makes sense. New test created for this purpose tied to `alter_audit_by _attempt`.

pyracf/resource/resource_admin.py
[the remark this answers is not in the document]
ElijahSwiftIBM: Thank you for the assist in refactoring. I am not always in the habit of thinking in format strings. This is all a lot cleaner. Changes made and the unit tests still pass which is good.

tests/resource/test_resource_getters.py
ElijahSwiftIBM: upon further investigation of audit rules i'm not sure it is possible to have a "different" permutation. I can discuss test results offline but it looks like auditing at each access level is not just success:update means audit all successful update accesses, it means audit all successes update and above. Because of this, I don't think its possible to have successes, failures, and all meaningfully defined to be 3 different rules, so RACF will only ever really have 2.

The logic is that for any ALL rule, it must either supersede at least one rule by being about a "lower" level of access, or be superseded by the combination of both rules which together encompass "all" auditing for 2 lower levels of access.

As such, no additional tests to add.

pyracf/resource/resource_admin.py
lcarcaramo: You are correct that this does only make the code marginally safer. You can correct me if I am wrong, but virtually anything in python is convertible to a string. `ThreadPoolExecutor` for example likely does not have a to string function, so I think Python just defaults to giving you a generic string telling you what kind of object the variable is and its address.

```python
>>> from concurrent.futures import ThreadPoolExecutor
>>> executor = ThreadPoolExecutor(max_workers=10)
>>> str(executor)
'<concurrent.futures.thread.ThreadPoolExecutor object at 0x000002532315DAC0>'
>>>
```

So, the only case that I can think of where there might be some strangeness, would  be if the user provided an object whose to string method returned a valid access level.

```python
>>> class MyClass():
...     def __str__(self):
...             return "alter"
...
>>> my_object = MyClass()
>>> str(my_object)
'alter'
```

And given that Python is in no way shape or form a type safe language, it can be argued that this is not a problem at all because this aligns with "duck typing", which is is not something that I don't think we have used in this project, but it is something that is fully support by Python and I am pretty sure it is probably recommended as a Pythonic way of doing things when it makes sense to do so. So from the perspective of `__validate_access_levels()` we could say in this case that given that `MyClass` looks like a string value of `alter` when we convert it to a string using `str()`, it is a valid access level. This more or less fits the "duck typing" paradigm: "If it looks like a duck and it quacks like a duck, then it is a duck."

So, in conclusion,  I agree that the benefit we get here is marginal, and I admit that this check may not be the most Pythonic, so after looking at this closer, I am in agreement that it may actually be better to just check if the value is `None` and when we convert the value to a string using `str()`, check if it ends up being a valid access level.

tests/resource/test_resource_setters.py
```diff
@@ -18,10 +18,278 @@ class TestResourceSetters(unittest.TestCase):
...
+        result = self.resource_admin.overwrite_audit_by_attempt(
+            "TESTING", "ELIJTEST", success="alter", all="read", failure="update"
+        )
+        self.assertEqual(
+            result,
+            TestResourceConstants.TEST_RESOURCE_OVERWRITE_AUDIT_BY_ATTEMPT_ALL_REQUEST_XML,
+        )
+
+    def test_resource_admin_build_overwrite_audit_by_attempt_value_error(self):
+        bad_val = "problem"
```
lcarcaramo, line 120: `bad_success = "problem"` instead of `bad_val = "problem"`.

pyracf/resource/resource_admin.py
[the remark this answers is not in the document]
lcarcaramo: My only suggestion now, would just be to further clean up your string building.

For example, I would change `case 1` to:

```python
            case 1:
                value_error_text = (
                    f"'{bad_access_levels[0]}' is not a valid access level. {value_error_text}"
                )
```

I think you can do something similar for your default case as well.

tests/resource/test_resource_getters.py
lcarcaramo: Maybe also the case where there are no audit rules as well.

tests/resource/test_resource_setters.py
```diff
@@ -18,10 +18,278 @@ class TestResourceSetters(unittest.TestCase):
...
+        with self.assertRaises(ValueError) as exception:
+            self.resource_admin.overwrite_audit_by_attempt(
+                "TESTING", "ELIJTEST", success=bad_val
+            )
+        print(exception.exception, str(exception.exception).encode("utf-8"))
```
lcarcaramo, line 125: Extraneous `print()`.

Also, is why is `exception.exception` being encoded here?

ElijahSwiftIBM: I honestly don't remember. I was running into something weird in testing and that was an attempt at debugging it. Also found another extraneous print that was removed as well. Should be all clear there.

tests/resource/test_resource_setters.py
ElijahSwiftIBM: very fair point. changed.

tests/resource/test_resource_getters.py
ElijahSwiftIBM: very fair point. I have added tests for that now.

pyracf/resource/resource_admin.py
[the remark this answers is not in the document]
ElijahSwiftIBM: default case is currently hamstrunf by the line length requirements at least in the following area (where I believe you are referring)

```python3
    value_error_text = (
                    f"{', '.join(bad_access_levels)}are not valid access levels. "
                    + f"{value_error_text}"
                )
```

works fine, but

```python3
                value_error_text = (
                    f"{', '.join(bad_access_levels)}are not valid access levels. {value_error_text}"
                )
```
leaves a line as 101 characters.

Your instructions for case 1 have been followed though. This was the natural progression of your other comments and it should have been something i did in the last commit.

pyracf/resource/resource_admin.py
ElijahSwiftIBM: Ok, that makes a lot of sense, and I think I do agree with the theory behind that. Definitely interesting to keep in mind for future work in the area. Based on this dicussion, I have changed it back to:

```python3
        for attempt_argument in (success, failure, all):
            if (
                attempt_argument is not None
                and str(attempt_argument).lower() not in valid_access_levels
            ):
                bad_access_levels.append(attempt_argument)
```

with this philosophy in mind.

pyracf/resource/resource_admin.py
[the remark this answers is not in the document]
ElijahSwiftIBM: Actually, case1 ended up in a similar situation:

```python3
                value_error_text = f"'{bad_access_levels[0]}' is not a valid access level. {value_error_text}"
```
The above implementation is too long

```python3
                value_error_text = (
                    f"'{bad_access_levels[0]}' is not a valid access level. {value_error_text}"
                )
```
I tried this to get around it but black formatter "optimizes" it into the previous implementation which flake8 then complains about

```python3
                value_error_text = (
                    f"'{bad_access_levels[0]}' is not a valid access level. "
                    + f"{value_error_text}"
                )
```

This then works through both, but in my opinion looks worse than either of the above. If there's an obvious way around this, I am missing it.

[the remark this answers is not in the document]
lcarcaramo: If this refactoring is putting you over the line limit, then I think you are good.

pyracf/resource/resource_admin.py
```diff
@@ -589,3 +719,67 @@ def _format_profile(self, result: dict) -> None:
...
+                ]
+                bad_access_levels[-1] = f"and {bad_access_levels[-1]} "
+                value_error_text = (
+                    f"{', '.join(bad_access_levels)}are not valid access levels. "
+                    + f"{value_error_text}"
+                )
+        raise ValueError(value_error_text)
+
+    def __check_for_duplicates(
+        self, input_list: list, argument: str, value: str
```
lcarcaramo, line 768: I don't think it is necessary to have `argument` and `value` be arguments here since `__validate_access_levels()` is the only caller and this is only called once with hard coded values.

pyracf/resource/resource_admin.py
```diff
@@ -589,3 +719,67 @@ def _format_profile(self, result: dict) -> None:
...
+    ) -> None:
+        duplicates = [
+            key
+            for (key, value) in Counter(input_list).items()
+            if (value > 1 and not (key is None))
+        ]
```
lcarcaramo, line 774: I think this should be:
```python
        duplicates = [
            key
            for (key, value) in Counter(input_list).items()
            if (value > 1 and key is not None)
        ]
```

pyracf/resource/resource_admin.py
lcarcaramo: Maybe change `input_list` to `argument_list` since we are checking arguments here?

pyracf/resource/resource_admin.py
```diff
@@ -589,3 +719,67 @@ def _format_profile(self, result: dict) -> None:
...
+        if duplicates == []:
+            return
+        value_error_text = []
+        for duplicate in duplicates:
+            value_error_text.append(
+                f"You entered '{duplicate}' for multiple parameters which is not supported."
```
lcarcaramo, line 780:
```python
f"'{duplicate}' is provided as an '{argument}' multiple times, which is not allowed."
```

pyracf/resource/resource_admin.py
```diff
@@ -589,3 +719,67 @@ def _format_profile(self, result: dict) -> None:
...
+            )
+        value_error_text.append(
+            f"Please use multiple function calls to set multiple {argument} to the same {value}."
```
lcarcaramo, line 783: I am not sure this message is necessary.

pyracf/resource/resource_admin.py
lcarcaramo: It may actually make sense to keep `argument` as an argument and when calling the function give it a value of `"Access Level"`.

pyracf/resource/resource_admin.py
ElijahSwiftIBM: Agree this is better.

pyracf/resource/resource_admin.py
ElijahSwiftIBM: With the above change in place, it is not. This has been removed.

pyracf/resource/resource_admin.py
ElijahSwiftIBM: yes, that is clearer. Changed.

pyracf/resource/resource_admin.py
ElijahSwiftIBM: I find the overall wording of this somewhat confusing, but I do still think it is better to be explicit when possible, so I have changed it as described.

tests/resource/test_resource_constants.py
```diff
@@ -111,6 +111,67 @@ def get_sample(sample_file: str) -> Union[str, bytes]:
...
+)
+
+# ============================================================================
+# Resource Administration Getters Result Data
+# ============================================================================
+
+TEST_EXTRACT_RESOURCE_GET_AUDIT_RULES = {"success": "update", "failures": "read"}
+TEST_EXTRACT_RESOURCE_GET_AUDIT_RULES_SINGLE = {"failures": "read"}
+
+TEST_EXTRACT_RESOURCE_GET_AUDIT_RULES_WITH_ALL = {"success": "update", "all": "read"}
```
lcarcaramo, line 174: The inclusion `EXTRACT_RESOURCE` might be a little confusing here. I would consider removing it from these samples.

There also appears to be an extraneous space here.

pyracf/resource/resource_admin.py
```diff
@@ -285,6 +286,133 @@ def get_user_access(
...
+        if success is not None:
+            traits[f"base:audit_{success}"] = "success"
+        if failure is not None:
+            traits[f"base:audit_{failure}"] = "failure"
+        if all is not None:
+            traits[f"base:audit_{all}"] = "all"
+        result = self.alter(resource, class_name, traits=traits)
+        return self._to_steps(result)
+
+    def alter_audit_by_attempt(
```
lcarcaramo, line 322: Does it make sense to update the naming convention on these to `audit_rules_by` instead of `audit_by`?

tests/resource/test_resource_constants.py
ElijahSwiftIBM: Sure, taken out.

pyracf/resource/resource_admin.py
ElijahSwiftIBM: This did make some test variables VERY long, but tests seem to still pass after what i've done to keep the lines short.

ElijahSwiftIBM: actually, black formatter undid my changes, so I'm not sure there's an appropriate way to keep some of these test constants within an appropriate length. I would have to either amend the naming scheme or make special consideration to the following test constants:

```
TestResourceConstants.TEST_RESOURCE_OVERWRITE_AUDIT_RULES_BY_ACCESS_LEVEL_NONE_REQUEST_XML,
TestResourceConstants.TEST_RESOURCE_OVERWRITE_AUDIT_RULES_BY_ACCESS_LEVEL_MULT_REQUEST_XML,
TestResourceConstants.TEST_RESOURCE_OVERWRITE_AUDIT_RULES_BY_ACCESS_LEVEL_ALL_REQUEST_XML,
```

I am alright making changes like shortening OVERWRITE to OVW for these constants and adding a comment in our constant definitions that would explain why.

ElijahSwiftIBM: Or i can leave in the exceptions.

lcarcaramo: I think we definitely want to resolve `E501`. I will mostly leave it up to how to shorten the variables. Unfortunately, we will need to change at least one word in the variable name to an acronym.

lcarcaramo: Maybe change `OVERWRITE` to `OVWT`?

lcarcaramo: Actually maybe you can even just remove the word `RULES`. I think that would be acceptable for a test variable.

lcarcaramo: This would be a rare exception to trying to make test constant names and test case name match the functions they are testing.

lcarcaramo (approved):

Issue #49, opened 2023-11-17
Feature Request: EXPIRED/NOEXPIRED option on set_password and set_passphrase


ElijahSwiftIBM:
**Is your feature request related to a problem? Please describe.**
set_password and set_passphrase options currently don't have the option to set non-expired passwords.

**Describe the solution you'd like**
We should add an optional argument to these functions to make the password non-expired.

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

lcarcaramo:
Resolved by #55

PR #55, opened 2023-11-24
Add optional Expired/Noexpired

ElijahSwiftIBM:
-Add optional parameter to set_password and set_passphrase that allows for user passwords and phrases to be set noexpired

### :bulb: Issue Reference

**Issue:**  #49 

Admins setting user passwords or phrases may want to make them non-expired, which is not possible except with UserAdmin.alter() right now.

Added an optional argument to set_password and set_passphrase that emulates RACF defaults anyway (setting a password without NOEXPIRED will make the password EXPIRED by default).

Made 2 new testcases to test setting noexpired with password and password phrase

### DIFF
--- a/pyracf/resource/resource_admin.py
+++ b/pyracf/resource/resource_admin.py
@@ -512,8 +512,8 @@ class ResourceAdmin(SecurityAdmin):
             return self._make_request(profile_request, irrsmo00_precheck=True)
         try:
             profile = self.extract(resource, class_name, profile_only=True)
-        except SecurityRequestError:
-            raise AlterOperationError(resource, class_name)
+        except SecurityRequestError as exception:
+            raise AlterOperationError(resource, class_name) from exception
         if not self._get_field(profile, "base", "name") == resource.lower():
             raise AlterOperationError(resource, class_name)
         self._build_segment_trait_dictionary(traits)
--- a/pyracf/user/user_admin.py
+++ b/pyracf/user/user_admin.py
@@ -275,24 +275,31 @@ class UserAdmin(SecurityAdmin):
     # Password
     # ============================================================================
     def set_password(
-        self,
-        userid: str,
-        password: Union[str, bool],
+        self, userid: str, password: Union[str, bool], expired: bool = True
     ) -> Union[dict, bytes]:
         """Set a user's password."""
-        result = self.alter(userid, traits={"base:password": password})
+        if not password:
+            result = self.alter(userid, traits={"base:password": password})
+            return self._to_steps(result)
+        result = self.alter(
+            userid, traits={"base:password": password, "base:password_expired": expired}
+        )
         return self._to_steps(result)
 
     # ============================================================================
     # Passphrase
     # ============================================================================
     def set_passphrase(
-        self,
-        userid: str,
-        passphrase: Union[str, bool],
+        self, userid: str, passphrase: Union[str, bool], expired: bool = True
     ) -> Union[dict, bytes]:
         """Set a user's passphrase."""
-        result = self.alter(userid, traits={"base:passphrase": passphrase})
+        if not passphrase:
+            result = self.alter(userid, traits={"base:passphrase": passphrase})
+            return self._to_steps(result)
+        result = self.alter(
+            userid,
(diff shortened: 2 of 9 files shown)


pyracf/user/user_admin.py
```diff
@@ -275,24 +275,31 @@ def take_away_auditor_authority(self, userid: str) -> Union[dict, bytes]:
     # Password
     # ============================================================================
     def set_password(
```
lcarcaramo, line 277: This might be a little cleaner, but I am not sure how a password set to `False` is intended to affect the logic here.

```python
def set_password(
Leave a comment
        self, userid: str, password: Union[str, bool], expired: bool = True
    ) -> Union[dict, bytes]:
        """Set a user's password."""
        traits = {"base:password": password}
        if expired:
            traits["base:password_expired"] = True
        return self.to_steps(self.alter(userid, traits=traits))
```

pyracf/resource/resource_admin.py
```diff
@@ -512,8 +512,8 @@ def alter(self, resource: str, class_name: str, traits: dict) -> Union[dict, byt
             return self._make_request(profile_request, irrsmo00_precheck=True)
         try:
             profile = self.extract(resource, class_name, profile_only=True)
-        except SecurityRequestError:
-            raise AlterOperationError(resource, class_name)
+        except SecurityRequestError as exception:
+            raise AlterOperationError(resource, class_name) from exception
```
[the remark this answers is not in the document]
ElijahSwiftIBM: So this was born out of an issue where pylint was acting up and reviewing files it had no business touching, but I did see some value in it. This attributes the AlterOperationError as being directly caused by the SecurityRequestError instead of just something that occurred while handling it. I found a [Stack Overflow thread that explains it nicely](https://stackoverflow.com/questions/24752395/python-raise-from-usage). I am okay deleting this, but I do think it might be better to keep around.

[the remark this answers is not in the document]
lcarcaramo: Okay, I understand. This just helps make it clear that `AlterOpertionError` was caused by `SecurityRequestError`.

pyracf/user/user_admin.py
lcarcaramo: Change `if password:` to `if expired` and rename `request_traits` to `traits` since this is consistent with naming used elsewhere. Do the same with `set_passphrase`.

lcarcaramo: So, you are saying that if `expired` is `False`, we want to explicitly set `base:password_expired` to `False`, otherwise we don't specify anything for `base:password_expired` which causes RACF's default behavior to take effect *(set the new password as expired by default)*?

lcarcaramo: Either way, I also think there should be tests for both `expired=True` and `expired=False`. The test for `expired=True` is a little redundant, but I think it would be a good regression test. If the default behavior differs from setting `expired=True` explicitly, then that is indicative of a problem.

ElijahSwiftIBM: That's a fair point. I changed things around a bit again. This is the code logic. As long as expired is specified explicitly, the value will be used. If it is not, it won't be. This should be functionally the same as it was, but this will be the cleanest way to deliver exactly what the user would want as a result.

```python3
    # Password
    def set_password(
        self, userid: str, password: Union[str, bool], expired: Union[bool, None] = None
    ) -> Union[dict, bytes]:
        """Set a user's password."""
        traits = {"base:password": password}
        if expired is not None:
            traits["base:password_expired"] = expired
        result = self.alter(userid, traits=traits)
        return self._to_steps(result)

    # Passphrase
    def set_passphrase(
        self, userid: str, passphrase: Union[str, bool], expired: Union[bool, None] = None
    ) -> Union[dict, bytes]:
        """Set a user's passphrase."""
        traits = {"base:passphrase": passphrase}
        if expired is not None:
            traits["base:password_expired"] = expired
        result = self.alter(userid, traits=traits)
        return self._to_steps(result)
```

I then added explicit test cases for password and password phrase with expired=True and the additional request xml files to match.

lcarcaramo (approved):

PR #53, opened 2023-11-24
Error and Error Handling Changes

ElijahSwiftIBM:
-Standardize SecurityRequestError Structure, merging IRRSMO00 Errors and that of RACF Errors with the smae internal structure in the result dictionary
-Add an install script that defines IRR.IRRSMO00.PRECHECK with UACC of none and/or checks if the profile exists and the current user's access -Added NullResponseError
-Added Unit Testing for Install script and null response error

### :bulb: 

**Issue:**  #44 #48 #52 

### :computer: What does this address?

Addresses listed issues in pyRACF's error handling and associated processing.
1) Adds new error for Null Response from IRRSMO00
2) Standardizes SecurityRequestError structure
3) Adds "install script" to define/check authorizations for IRR.IRRSMO00.PRECHECK

### :pager: Implementation Details

1) Checked for null string response from IRRSMO00 and throw new error if so
2) Define new methods of SecurityRequestError to restructure IRRSMO00 response XML with IRRSMO00 error to more closely align with RACF-error structure
3) Added "Scripts" folder and defined externally available function that checks active user's access to IRR.IRRSMO00.PRECHECK and creates the resource if possible.

### :clipboard: Is there a test case?

Designed new test cases for new error and install script under "common" test cases. Tested new SecurityRequestError functions with existing user and group error tests

Issue #48, opened 2023-11-17
Feature Request: Install Script/Documentation Improvements


ElijahSwiftIBM:
**Is your feature request related to a problem? Please describe.**
Users are having issues with the setup and definition of the `IRR.IRRSMO00.PRECHECK` resource in the `XFACILIT` class. 

**Describe the solution you'd like**
We should potentially have a script that could define this resource so that auditing would flag attempts to use it when it fails. 

**Describe alternatives you've considered**
Another alternative would be to DIRECTLY REFERENCE this general resource in the install steps in the documentation.

**Additional context**
N/A

lcarcaramo:
Implemented in #57

Issue #45, opened 2023-11-16
Feature Request: Better Error Handling for Empty Return from IRRSMO00


ElijahSwiftIBM:
**Is your feature request related to a problem? Please describe.**
Users are starting to use pyracf without ensuring all setup steps are complete like the IRR.IRRSMO00.PRECHECK profile that we document. Without this profile, alter profile requests result in empty return values from IRRSMO00, which pyracf does not handle gracefully as shown:

```
  File "/usr/lpp/IBM/cyp/v3r10/pyz/lib/python3.10/site-packages/pyracf/user/user_admin.py", line 295, in set_passphrase
    result = self.alter(userid, traits={"base:passphrase": passphrase})
  File "/usr/lpp/IBM/cyp/v3r10/pyz/lib/python3.10/site-packages/pyracf/user/user_admin.py", line 801, in alter
    return self._make_request(user_request, irrsmo00_precheck=True)
  File "/usr/lpp/IBM/cyp/v3r10/pyz/lib/python3.10/site-packages/pyracf/common/security_admin.py", line 178, in _make_request
    results = SecurityResult(result_xml)
  File "/usr/lpp/IBM/cyp/v3r10/pyz/lib/python3.10/site-packages/pyracf/common/security_result.py", line 13, in __init__
    self.__result = XMLParser.fromstring(result_xml)
  File "/usr/lpp/IBM/cyp/v3r10/pyz/lib/python3.10/site-packages/defusedxml/common.py", line 127, in fromstring
    return parser.close()
  File "/usr/lpp/IBM/cyp/v3r10/pyz/lib/python3.10/xml/etree/ElementTree.py", line 1724, in close
    self._raiseerror(v)
  File "/usr/lpp/IBM/cyp/v3r10/pyz/lib/python3.10/xml/etree/ElementTree.py", line 1624, in _raiseerror
    raise err
xml.etree.ElementTree.ParseError: no element found: line 1, column 0
```

**Describe the solution you'd like**
This should result in a new error that should also steer users to this documentation to correct this mistake on their own.

**Describe alternatives you've considered**
Writing errors this severe to a log file of some sort could also be beneficial.

**Additional context**
N/A

lcarcaramo:
Resolved in #57

Issue #44, opened 2023-11-16
Feature Request: SecurityRequestError Structure Standardization


ElijahSwiftIBM:
**Is your feature request related to a problem? Please describe.**
When you run into "error" xml returned from IRRSMO00, there are effectively two kinds of structures:

Structure 1 is the case we based SecurityResultError on, RACF returns with a non-zero RC and the following XML:

```
<?xml version="1.0" encoding="IBM-1047"?>
<securityresult xmlns="http://www.ibm.com/systems/zos/saf/IRRSMO00Result1">
  <user name="TESTBOY" operation="set" requestid="UserRequest">
    <info>Definition exists. Add command skipped due  to precheck option</info>
    <command>
      <safreturncode>8</safreturncode>
      <returncode>16</returncode>
      <reasoncode>8</reasoncode>
      <image>ALTUSER TESTBOY  OMVS     (UID         (STRING))</image>
      <message>IKJ56701I MISSING OMVS UID+</message>
      <message>IKJ56701I MISSING OMVS USER ID (UID), 1-10 NUMERIC DIGITS</message>
      <message>IKJ56716I EXTRANEOUS INFORMATION WAS IGNORED: STRING</message>
    </command>
  </user>
  <returncode>4</returncode>
  <reasoncode>0</reasoncode>
</securityresult>
```

Structure 2 is the case where IRRSMO00 itself is returning an error to the user as XML:

```
<?xml version="1.0" encoding="IBM-1047"?>
<securityresult xmlns="http://www.ibm.com/systems/zos/saf/IRRSMO00Result1">
  <user name="TESTBOY" operation="set" requestid="UserRequest">
    <error>
      <errorfunction>10</errorfunction>
      <errorcode>2000</errorcode>
      <errorreason>76</errorreason>
      <errormessage>Data may not be specified for a boolean field.</errormessage>
      <erroroffset>216</erroroffset>
      <textinerror>STRING</textinerror>
    </error>
  </user>
  <returncode>2000</returncode>
  <reasoncode>76</reasoncode>
</securityresult>
```

Because the internal XML structure is so different, the SecurityRequestError is very different. Our documented sample of:

```
from pyracf import UserAdmin
from pyracf import SecurityRequestError

user_admin = UserAdmin()

try:
    user_admin.alter("squidwrd", traits={"base:password": "passwordtoolong"})
except SecurityRequestError as e:
    return_code = e.result["securityResult"]["user"]["returnCode"]
[...]

PR #57, opened 2023-11-26
Run as UserID and Error handling changes

ElijahSwiftIBM:
### :bulb: Issue Reference

**Issue:**  #44 #48 #45 #43 

Addresses listed issues in pyRACF's error handling and associated processing.

1. Adds new error for Null Response from IRRSMO00
2. Standardizes SecurityRequestError structure
3. Adds "install script" to define/check authorizations for IRR.IRRSMO00.PRECHECK

Adds entirely new feature to run commands as another user. This requires specific RACF authorizations.

1. Checked for null string response from IRRSMO00 and throw new error if so
2. Define new methods of SecurityRequestError to restructure IRRSMO00 response XML with IRRSMO00 error to more closely align with RACF-error structure
3. Added "Scripts" folder and defined externally available function that checks active user's access to IRR.IRRSMO00.PRECHECK and creates the resource if possible.

Changed call to IRRSMO00 to pass in optional userid parameter which is set at the SecurityAdmin structure level. This can be set at object creation or with public access methods. The "running userid" when it exists is added to the securityResult dictionary for logging/tracking purposes.

Also changed call to IRRSMO00 to return more information than just xml string. With full return and reason codes, NullResponseErrors can be differentiated by their causes. This is also why this PR contains code from #53 and makes it redundant.

Designed new test cases for new error, install script, and run as userid functions under "common" test cases. Tested new SecurityRequestError functions with existing user and group error tests


pyracf/common/improper_userid_error.py
```diff
@@ -0,0 +1,17 @@
+"""Exception to use when attempting to set an invalid userid for pyRACF to run under."""
+
+
+class ImproperUserIDError(Exception):
```
lcarcaramo, line 4: I suggest changing the name of this exception to `UserIdError`.

As for making the letter 'd' in 'ID' lower case, there seems to be some disagreement in the community on how acronyms and more specifically two letter acronyms should be handled in camel/pascal case. Google's JavaScript Style Guide and a text to pascal case converter that I found, seem to indicate that all acronyms should follow the pascal case conventions like all other words. I am inclined to go with that standard so that there is no confusion.

https://google.github.io/styleguide/jsguide.html#naming-camel-case-defined
https://www.atqed.com/string-pascal-case

pyracf/common/irrsmo00.py
```diff
@@ -19,9 +20,29 @@ def __init__(self) -> None:
         # Initialize size of output buffer
         self.buffer_size = 100000
 
-    def call_racf(self, request_xml: bytes, precheck: bool = False) -> str:
+    def call_racf(
```
lcarcaramo, line 23: Suggested refactoring. This should be functionally the same, but just a little more concise and cleaner.

```
    def call_racf(
        self,
        request_xml: bytes,
        precheck: bool = False,
        run_as_userid: Union[str, None] = None,
    ) -> str:
        """Make request to call_irrsmo00 in the cpyracf Python extension."""
        options = 15 if precheck else 13
        userid = b""
        userid_length = 0
        if run_as_userid:
            userid = run_as_userid.encode("cp1047")
            userid_length = len(run_as_userid)
        response = call_irrsmo00(
            xml_str=request_xml,
            xml_len=len(request_xml),
            opts=options,
            userid=userid,
            userid_len=userid_length
        ).decode("cp1047")
        if response[0] == "":
            return response[1:3]
        return response[0]
```

pyracf/common/improper_userid_error.py
```diff
@@ -0,0 +1,17 @@
...
+    """
+    Raised when pyRACF would attempt to establish itself to
+    run under a passed 'userid' that is not a string of 1 to 8 characters.
+    """
+
+    def __init__(self, userid: str) -> None:
+        self.message = (
+            "Cannot run under this userid."
+            + f"{userid} is not a string from 1 to 8 characters in length."
+        )
```
lcarcaramo, line 14:
```python
    """
    Raised when pyRACF would attempt to run as a userid
    that is not a string between 1 to 8 characters in length.
    """

    def __init__(self, userid: str) -> None:
        self.message = (
            "Unable to run as userid '{userid}'. Userid must "
            + f"be a string value between 1 to 8 characters in length."
        )
```

pyracf/common/null_response_error.py
```diff
@@ -0,0 +1,45 @@
+"""Exception to use when no data is returned by IRRSMO00."""
+from typing import Union
+
+
+class NullResponseError(Exception):
+    """
+    Raised when no xml string is returned by IRRSMO00.
+    """
+
+    def __init__(self, xml_str: str, run_as_userid: Union[str, None] = None) -> None:
```
lcarcaramo, line 10: Will this exception handler print out the return and reason codes for you? I know that they are stored as attributes on the exception handler, but if someone's code doesn't handle this exception and inspect the return and reason codes, the user may never know what happened. This exception should probably always print the return and reason code to facilitate debugging especially in the case where the the return and reason code combination is not one known by pyRACF.

pyracf/common/null_response_error.py
```diff
@@ -0,0 +1,45 @@
...
+        self.saf_return_code = xml_str[0]
+        self.racf_return_code = xml_str[1]
+        self.racf_reason_code = xml_str[2]
+        if (
+            (self.saf_return_code == 8)
+            and (self.racf_return_code == 200)
+            and (self.racf_reason_code == 16)
+        ):
+            self.message += (
+                "\n\nCheck to see if proper RACF permissions are in place.\n"
```
lcarcaramo, line 21: `the proper`

pyracf/common/null_response_error.py
```diff
@@ -0,0 +1,45 @@
...
+                + "For `set` or `alter` functions, you must have at least READ "
+                + "access to `IRR.IRRSMO00.PRECHECK` in the `XFACILIT` class."
+            )
+        elif (
+            (self.saf_return_code == 8)
+            and (self.racf_return_code == 200)
+            and (self.racf_reason_code == 8)
+        ):
+            self.message += (
+                "\n\nCheck to see if proper RACF permissions are in place.\n"
```
lcarcaramo, line 31: `the proper`

pyracf/common/improper_userid_error.py
ElijahSwiftIBM: That is a reasonable direction to go with it. I will make sure that is changed in the necessary spots.

pyracf/common/improper_userid_error.py
ElijahSwiftIBM: change made.

pyracf/common/null_response_error.py
ElijahSwiftIBM: changed

pyracf/common/null_response_error.py
ElijahSwiftIBM: changed

pyracf/common/improper_userid_error.py
lcarcaramo: Can you also change `ImproperUserIdError` to `UserIdError`? I don't know if the token `Improper` is necessary. I think `UserIdError` gets the point across.

pyracf/common/null_response_error.py
lcarcaramo: I think it would be good for item to be printed on a separate line.

```console

SAF Return Code: 8
RACF Return Code: 200
RACF Reason Code: 16

Check to see if the proper RACF permissions are in place.
For `set` or `alter` functions, you must have at least READ access to `IRR.IRRSMO00.PRECHECK` in the `XFACILIT` class.
```

pyracf/common/security_admin.py
```diff
@@ -20,6 +22,7 @@ class SecurityAdmin:
     _valid_segment_traits = {}
     _extracted_key_value_pair_segment_traits_map = {}
     _case_sensitive_extracted_values = []
+    __running_userid = False
```
lcarcaramo, line 25: This should default to `None`, no?

pyracf/common/security_request_error.py
```diff
@@ -18,9 +19,46 @@ def __init__(self, result: dict) -> None:
     def __str__(self) -> str:
         return self.message
 
+    def __check_result_type(self, result: dict) -> dict:
```
lcarcaramo, line 22: Can we move this logic to `security_result.py`? All of that transformation of the result xml into a result dictionary should be handled in `security_result.py` with the exception of profile extract since those are all function group specific.

pyracf/common/security_result.py
```diff
@@ -9,10 +10,14 @@
...
 
-    def __init__(self, result_xml: str) -> None:
+    def __init__(
+        self, result_xml: str, running_userid: Union[str, None] = None
+    ) -> None:
         self.__result = XMLParser.fromstring(result_xml)
         self.__result_dictionary = {"securityResult": {}}
         self.__extract_results()
+        if running_userid is not None:
+            self.__result_dictionary["securityResult"]["runningUserid"] = running_userid
```
lcarcaramo, line 20: Does it make sense to have it default to the current user for audit purposes? You can use `getpass` to do this in Python.

https://stackoverflow.com/questions/842059/is-there-a-portable-way-to-get-the-current-username-in-python

tests/common/test_install_precheck_script.py
```diff
@@ -0,0 +1,157 @@
+"""Test password sanitization in resource debug logging."""
```
lcarcaramo, line 1: I don't think this docstring is correct.

tests/group/group_result_samples/add_group_result_error.json
```diff
@@ -4,14 +4,17 @@
...
-        "errorMessage": "Invalid attribute value specified.",
-        "errorOffset": 149,
-        "textInError": "name"
-      }
+      "commands": [
+        {
+          "safReturnCode": 10,
+          "returnCode": 2000,
+          "reasonCode": 68,
+          "image": "<securityrequest xmlns=\"http://www.ibm.com/systems/zos/saf\" xmlns:racf=\"http://www.ibm.com/systems/zos/racf\"><group name=\"TESTGRPP0\" operation=\"set\" requestid=\"GroupRequest\" /></securityrequest>",
```
lcarcaramo, line 12: Why does "image" contain an xml string? If we don't have a valid value for "image" from the original response maybe we should just set this to `None`?

tests/group/group_result_samples/add_group_result_error.json
```diff
@@ -4,14 +4,17 @@
...
+          "messages": [
+            "Invalid attribute value specified.\nText in Error: 'name'\nApproximate Offset of Text in Error: '149'\nPlease note that for any text in error, redacted values may skew offset calculations."
```
lcarcaramo, line 14: I think these messages need to be split into separate lines:

```json
          "messages": [
            "Invalid attribute value specified.",
            "Text in Error: 'name'",
            "Approximate Offset of Text in Error: '149'",
            "Please note that for any text in error, redacted values may skew offset calculations."
          ]
```

Also, I don't know if it makes sense to have the xml elements as additional json keys like the old format so that it is easier to parse maybe:

```json
      "commands": [
        {
          "safReturnCode": 10,
          "returnCode": 2000,
          "reasonCode": 68,
          "image": ...,
          "messages": [
            "Invalid attribute value specified.",
            "Please note that for any text in error, redacted values may skew offset calculations."
          ]
          "errorOffset": 149,
          "textInError": "name"
        }
      ]
```

tests/group/test_group_result_parser.py
```diff
@@ -75,6 +75,7 @@ def test_group_admin_can_parse_add_group_error_xml(
             self.group_admin.add(
                 "TESTGRPP0", traits=TestGroupConstants.TEST_ADD_GROUP_REQUEST_TRAITS
             )
+        print(exception.exception.result)
```
lcarcaramo, line 78: Delete print statement.

tests/resource/resource_result_samples/extract_resource_result_precheck_error.xml
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="IBM-1047"?>
+<securityresult xmlns="http://www.ibm.com/systems/zos/saf/IRRSMO00Result1">
+  <resource name="IRR.IRRSMO00.PRECHECK" class="XFACILIT" operation="listdata" requestid="ResourceRequest">
+    <command>
+      <safreturncode>8</safreturncode>
+      <returncode>16</returncode>
+      <reasoncode>4</reasoncode>
+      <image>RLIST   XFACILIT             (IRR.IRRSMO00.PRECHECK) </image>
+      <message>ICH13003I TESTING NOT FOUND</message>
```
lcarcaramo, line 9: Is this the actual message that this returns?

pyracf/common/security_result.py
lcarcaramo: Also, I don't know if it makes sense to change `runningUserid` to something like `authorityUsed` so it is clear that it is saying that the following userid is the one who's authority was used to make the request?

pyracf/common/improper_userid_error.py
ElijahSwiftIBM: Fair point; done!

pyracf/common/null_response_error.py
ElijahSwiftIBM: That also makes sense; done!

pyracf/common/security_admin.py
ElijahSwiftIBM: Yes, good catch. I had originally established this with `False` having a role in this process and this is a relic that was missed in the transition to `None`. Changed.

pyracf/common/security_request_error.py
ElijahSwiftIBM: That's a very fair point, I moved all of it over there as this is really about the security_result structure.

pyracf/common/security_result.py
ElijahSwiftIBM: It could cause problems with RACF permissions if we were to force pyRACF into utilizing running_userid when it would not be intended. We could get around this with more checks on the userid, but that feels unnecessary.Our "audit purposes" would really only add the running userid to debug logging or structure output which I don't know really helps. 

I could consider a name change to authorityUsed, but I like the idea of keeping it similar to IRRSMO00's RunAsUserID so that we can reference their documentation for the authority needed and have that make sense. We can discuss in more detail.

tests/common/test_install_precheck_script.py
ElijahSwiftIBM: Good catch. Changed to:
`"""Test included script to initialize and evaluate precheck permissions."""`

tests/group/group_result_samples/add_group_result_error.json
ElijahSwiftIBM: I made the intentional choice to put the request_xml there for IRRSMO00 errors. This is because 1) there is no technical command image for these as no command image was generated, IRRSMO00 returned before we got that far and 2) these errors often reference specific offsets into the input xml string where there are concerns, so I wanted a way to tether the input xml to the error. I could be open to changing this to its own key in these error dictionaries.

tests/group/group_result_samples/add_group_result_error.json
ElijahSwiftIBM: I could be open to leaving these in json keys rather than the message, or maybe both. I kept them in messages to standardize things as RACF's "errors" will surface them in MESSAGES rather than something as simple to parse as this. Open to further discussion. Once we have a defined solution, I can change this to do separate messages.

tests/group/test_group_result_parser.py
ElijahSwiftIBM: good catch; gone now.

tests/resource/resource_result_samples/extract_resource_result_precheck_error.xml
ElijahSwiftIBM: No, it should have been `<message>ICH13003I IRR.IRRSMO00.PRECHECK NOT FOUND</message>` and has now been changed.

pyracf/common/null_response_error.py
```diff
@@ -0,0 +1,49 @@
...
+                "\n\nCheck to see if the proper RACF permissions are in place.\n"
+                + "For the `run_as_userid` feature, you must have at least UPDATE"
+                + f" access to `{run_as_userid}.IRRSMO00` in the `SURROGAT` class."
+            )
+        else:
+            self.message += (
+                "\n\nPlease check the specified return and reason codes against"
+                + " the IRRSMO00 documented return and reason codes for more information"
+                + " about this error.\n"
+                + "(https://www.ibm.com/docs/en/zos/3.1.0?topic=operations-return-reason-codes)"
```
lcarcaramo, line 44: I don't think you need to surround the link with parenthesis here.

pyracf/common/security_result.py
```diff
@@ -56,6 +63,42 @@ def __extract_info(self) -> None:
...
+    def __check_result_type(self, result: dict) -> dict:
+        not_profiles = ["returnCode", "reasonCode", "runningUserid"]
+        for profile_type in result["securityResult"]:
+            if profile_type in not_profiles:
+                continue
+            if "error" in result["securityResult"][profile_type]:
+                return self.__organize_smo_error(result, profile_type)
+        return result
+
+    def __organize_smo_error(self, result: dict, profile_type: str) -> dict:
```
lcarcaramo, line 75: Can we change the name of this to `__standardize_error_response`? It could be good to include a short docstring to briefly explain why this post processing is needed.

lcarcaramo: Also, it probably makes sense to do this earlier in `SecurityResult.__extract_error`, no? Then there doesn't need to be a special extra post processing step. All of this can more or less just be handled in `SecurityResult.__extract_error`.

pyracf/common/security_result.py
lcarcaramo: I am okay with keeping `runningUserid`. However, my suggestion is not to always run in run as userid mode. my suggestion is that if `running_userid` is not set, then it is assumed that the current user running the program is the "running userid". Then we can just use get pass to get that userid and add it to the result dictionary:

```python
        if running_userid is None:
            self.__result_dictionary["securityResult"]["runningUserid"] = getpass.getuser()
        else:
            self.__result_dictionary["securityResult"]["runningUserid"] = running_userid
```

Also it is probably a good idea to standardize to lowercase for the values in the security result dictionary:

```python
self.__result_dictionary["securityResult"]["runningUserid"] = running_userid.lower()
```

pyracf/scripts/install_precheck.py
```diff
@@ -0,0 +1,34 @@
...
+            + "\nPlease ensure other users of pyRACF also have at least read access."
+            + "\nReview our documentation at https://ambitus.github.io/pyracf/ as well!"
+        )
+        return True
+    print(
+        "IRR.IRRSMO00.PRECHECK is already defined, but you have no access."
+        + "\nContact your security administrator for READ access before using pyRACF."
+        + "\nReview our documentation at https://ambitus.github.io/pyracf/ as well!"
+    )
+    return False
```
lcarcaramo, line 34: Just to confirm that I understand how this works. This code does the following:

1. get the user's current access to `IRR.IRRSMO00.PRECHECK`.
2. If `SecurityRequestError` is raised _(because precheck isn't set up?)_, try to define precheck with universal access of `None`.
3. If the user has access, they can use precheck.
4. If precheck is setup but the user doesn't have access, they have to contact a system programmer/security administrator.
5. My guess is that if adding the precheck profile fails, it is just supposed to let the `SecurityRequestError` bubble up.

pyracf/common/null_response_error.py
ElijahSwiftIBM: I agree, that looks better. Good call on the spaces, I had missed those. I do think there could be a benefit, but I doubt it would be as useful with those errors. I included them here because it can assist in troubleshooting especially for the "generic" case of a null response due to an unanticipated failure and not the permissions checks we are aware of. With SecurityRequestError, especially in its unified state, the "messages" list really gives the specific context about the failures without the need to directly reference documentation for the meanings of the codes.

pyracf/common/security_result.py
ElijahSwiftIBM: That would work functionally but based on the current call to RACF, this would drive IRRSMO00 to try and pass that userid as RunAsUserID which I could see as an issue. We could set up a check against the currently running user, or some sort of other flag involved in "clearing" or "setting" this value, but I am on the fence about whether that's necessary for this. I did cast it to lowercase though.

ElijahSwiftIBM: Oh, I just realized where you were proposing the codechange. I get it. I can look at that change.

tests/group/group_result_samples/add_group_result_error.json
ElijahSwiftIBM: Alright, that makes sense. Change made (including changes to NullResponseError).

tests/group/group_result_samples/add_group_result_error.json
ElijahSwiftIBM: Makes sense, change made.

pyracf/common/null_response_error.py
ElijahSwiftIBM: fair enough; change made.

pyracf/common/userid_error.py
```diff
@@ -0,0 +1,17 @@
+"""Exception to use when attempting to set an invalid userid for pyRACF to run under."""
```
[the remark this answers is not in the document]
ElijahSwiftIBM: Agree; change made.

pyracf/common/security_result.py
ElijahSwiftIBM: The more I think about this, the more I think that smo_error should be lumped in with null_response_error rather than security_request_error and both their names should be changed.

These error objects that have been "standardized" to look like result dictionaries SHOULD have a different layout. There are no commands that were generated or run, more information may be available in the IRRSMO00 return and reason codes, and it makes more sense for that to be a SecurityRequestError because that means there was an error with the fields or authorities to create and/or run the security request that we sent to IRRSMO00.

What we have been using as the "standard" for SecurityRequestError is REALLY IRRSMO00 executed our security request successfully, but RACF issued some warnings or errors based on the commands generated from the Request (I thought about calling this SecurityResultError).

In short, you are right that it should be handed that way if we are putting these together as one error, but I no longer think that should be the case.

ElijahSwiftIBM: I avoided that in the past because it also would impose a lot of changes to unit testing, but I think it is necessary.

ElijahSwiftIBM: I figured out a way to do it that didn't break much of anything so I have now committed my proposed change to the PR.

pyracf/scripts/install_precheck.py
ElijahSwiftIBM: That is correct, there's no explicit handling of any errors in the script except the initial error on step 1 that leads into step 2. If step 1 fails for any other reason (only thing i can think of is they lack access to view information on the profile), then step 2 will as well and surfacing that error is fine.

If the user has any non-NONE access then they have at least READ authority, so that is why the check says they can use precheck. Defining the profile also does mention that an admin should be contacted as well.

pyracf/common/security_result.py
```diff
@@ -130,6 +143,12 @@ def __to_pascal_case(self, key: str) -> str:
             case _:
                 return key
 
+    def __get_current_userid(self) -> str:
+        if platform.system() == "OS/390":
+            return getpass.getuser().lower()
+        else:
+            return "testuser"
+
```
lcarcaramo, line 151: I think it is better to create a mock in the unit tests to do this instead of adding this to the pyracf source code.

I think you should be able to do something like this in `test_running.py`.

```python
@patch('getpass.getuser')
def __test_suite(get_user_mock: Mock) -> unittest.TestSuite:
    """Load and run each unit test class."""
    get_user_mock.return_value = "patrick"
    test_suite = unittest.TestSuite()
    test_loader = unittest.TestLoader()
```

I think that mock should then be inherited by all of the test suites.

tests/user/user_result_samples/extract_user_result_bad_attribute_error.json
```diff
@@ -8,12 +8,13 @@
         "errorFunction": 10,
         "errorCode": 2000,
         "errorReason": 68,
-        "errorMessage": "Invalid attribute value specified.",
+        "textInError": "name",
         "errorOffset": 149,
-        "textInError": "name"
+        "errorMessage": "Invalid attribute value specified."
```
lcarcaramo, line 13: Is there a reason why the ordering was changed here?

ElijahSwiftIBM: This is how you the error dictionary is/was built. This comes from IRRSMO00 rather than RACF, so this is coming from the "new" SecurityRequestError (previously NullResponseError) and this dictionary is truer to the error reported by SMO as this isn't associated with a "command", is not expected to have messages outside of one errorMessage, and uses a slightly different tag scheme for the same thing.

pyracf/common/security_result.py
ElijahSwiftIBM: That's fair, I avoided this because I thought it would be a lot to handle for a feature that could never be used outside of z/OS anyway (it's not in the request builder functionality), but I can try to make this adjustment.

pyracf/scripts/install_precheck.py
```diff
@@ -0,0 +1,34 @@
+from pyracf import ResourceAdmin, SecurityResponseError
+
+
+def define_precheck_profile():
```
lcarcaramo, line 4: Can we rename this script to `setup_precheck.py` and change the name of the function to `setup_precheck()`?

pyracf/common/security_result.py
lcarcaramo: I generally agree with this change. The only thing I am disagreement with is the naming. `SecurityRequestError` is already established as the exception that gets raised when a "security request" fails so that shouldn't change. So, when no request is able to be made we need a name that indicates that no request was able to be made. If I understand correctly, these errors occur as a result of IRROSM00's input validation. So maybe we can call this exception something like `DownstreamInputValidationError` to indicate that it was not pyRACF's input validation that failed, which we already have exceptions for, but instead it was IRRSMO00's inputs validation that happens downstream that fails. Let me know what you think.

lcarcaramo: That then also leaves the naming for when we get a "null" response from IRRSMO00. That exception should maybe be called something like `DownstreamFatalError` to indicate that something unrelated to the user's input or the pyRACF installation is preventing the request from being processed.

pyracf/common/security_result.py
lcarcaramo: It is generally good to avoid having code that is only ever used for testing in the source code. The only exception in our case is the encoding stuff, which I think is okay.

tests/user/user_result_samples/extract_user_result_bad_attribute_error.json
lcarcaramo: Ok, so you just copied and pasted what pyRACF/IRRSMO00 gave you?

pyracf/common/security_result.py
ElijahSwiftIBM: I am having trouble getting it to it to inherit this sort of mock case. We'll have to discuss next week.

pyracf/common/security_admin.py
```diff
@@ -79,11 +83,34 @@ def __init__(
...
+            self.__running_userid = new_userid
+            return
+        if (
+            isinstance(new_userid, str)
+            and (len(new_userid) <= 8)
+            and (not new_userid == "")
+        ):
+            self.__running_userid = new_userid.upper()
+            return
+        raise UserIdError(new_userid)
```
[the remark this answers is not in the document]
ElijahSwiftIBM: They will be pushed when the unit test fix is also addressed.

pyracf/common/security_result.py
ElijahSwiftIBM: I could not get patch to inherit the way we wanted, but forcing     `getpass.getuser = Mock(return_value="testuser")` in the test_suite worked well.

pyracf/common/downstream_fatal_error.py
```diff
@@ -0,0 +1,66 @@
+"""Exception to use when no data is returned by IRRSMO00."""
+from typing import Union
+
+
+class DownstreamFatalError(Exception):
+    """
+    Raised when no xml string is returned by IRRSMO00.
+    """
```
lcarcaramo, line 8: Update docstrings.

pyracf/common/security_admin.py
```diff
@@ -157,32 +179,57 @@ def _make_request(
...
-        results = SecurityResult(result_xml)
+        results = SecurityResult(result_xml, self.get_running_userid())
         if self.__debug:
             # No need to redact anything here since the result dictionary
             # already has secrets redacted when it is built.
             self.__logger.log_dictionary(
                 "Result Dictionary", results.get_result_dictionary()
             )
         result_dictionary = results.get_result_dictionary()
+        if result_dictionary["securityResult"]["returnCode"] > 4:
```
lcarcaramo, line 217:
```python
if result_dictionary["securityResult"]["returnCode"] == 8:
```

pyracf/common/downstream_fatal_error.py
```diff
@@ -0,0 +1,66 @@
...
+        ):
+            self.message += (
+                "\n\nCheck to see if the proper RACF permissions are in place.\n"
+                + "For the `run_as_userid` feature, you must have at least UPDATE"
+                + f" access to `{run_as_userid}.IRRSMO00` in the `SURROGAT` class."
+            )
+        else:
+            self.message += (
+                "\n\nPlease check the specified return and reason codes against"
+                + " the IRRSMO00 documented return and reason codes for more information"
```
lcarcaramo, line 59: `documentation`

pyracf/common/security_admin.py
```diff
@@ -79,11 +83,29 @@ def __init__(
...
+    def set_running_userid(self, new_userid: Union[str, None]) -> None:
+        if new_userid is None:
+            self.__running_userid = new_userid
+            return
+        if not isinstance(new_userid, str) or len(new_userid) > 8 or new_userid == "":
+            raise UserIdError(new_userid)
+        self.__running_userid = new_userid.upper()
+
+    def clear_running_userid(self) -> None:
+        self.__running_userid = None
```
lcarcaramo, line 100: Do we need a `clear_running_userid` function or is it good enough to just document that `set_running_userid` can take a `None` for the userid to indicate that it is being unset?

pyracf/common/downstream_fatal_error.py
```diff
@@ -0,0 +1,66 @@
...
+
+
+class DownstreamFatalError(Exception):
+    """
+    Raised IRRSMO00 returns with a return code of 8, indicating an error occured prior to request.
+    """
+
+    def __init__(
+        self,
+        xml_str: str,
```
lcarcaramo, line 12: `xml_str` is a list of integers, no? In that case this should be called `return_reason_codes: List[int]` instead.

pyracf/common/downstream_fatal_error.py
```diff
@@ -0,0 +1,66 @@
...
+        request_xml: bytes,
+        run_as_userid: Union[str, None] = None,
+        result_dict: dict = None,
```
lcarcaramo, line 15: Change `result_dict` to `result_dictionary`.

pyracf/scripts/setup_precheck.py
```diff
@@ -0,0 +1,34 @@
...
+        result = resource_admin.add(
+            "IRR.IRRSMO00.PRECHECK", "XFACILIT", traits=traits_precheck
+        )
+        print(
+            "IRR.IRRSMO00.PRECHECK is now defined with a `Universal Access` of None."
+            + "\nContact your security administrator for READ access before using pyRACF."
+            + "\nOther users of pyRACF will also need to have at least read access."
+            + "\nYou may also need to REFRESH the `XFACILIT` class."
+            + "\nReview our documentation at https://ambitus.github.io/pyracf/ as well!"
+        )
```
lcarcaramo, line 19: `\n` should go at the end of the previous line instead of at the beginning of the current line.

```python
        print(
            "IRR.IRRSMO00.PRECHECK is now defined with a `Universal Access` of None.\n"
            + "Contact your security administrator for READ access before using pyRACF.\n"
            + "Other users of pyRACF will also need to have at least read access.\n"
            + "You may also need to REFRESH the `XFACILIT` class.\n"
            + "Review our documentation at https://ambitus.github.io/pyracf/ as well!"
        )
```

tests/common/test_common_constants.py
```diff
@@ -0,0 +1,143 @@
...
+import tests.test_utilities as TestUtilities
+
+
+def get_sample(sample_file: str, function_group: str) -> Union[str, bytes]:
+    return TestUtilities.get_sample(sample_file, function_group)
+
+
+# ============================================================================
+# Install Precheck Sample Data
+# ============================================================================
```
lcarcaramo, line 16:
```python
# Setup Precheck Sample Data
```

tests/common/test_common_constants.py
```diff
@@ -0,0 +1,143 @@
...
+TEST_ALTER_USER_SUCCESS_AS_ESWIFT_LOG = get_sample(
+    "alter_user_success_as_eswift.log", "user"
+)
+TEST_ALTER_USER_RESULT_SUCCESS_XML = get_sample("alter_user_result_success.xml", "user")
+TEST_EXTRACT_RESOURCE_PRECHECK_AS_SQUIDWRD_LOG = get_sample(
+    "extract_resource_precheck_as_squidwrd.log", "resource"
+)
+
+# ============================================================================
+# Null Response Error
```
lcarcaramo, line 117: This should reference `DownstreamFatalError` now, no?

tests/common/test_security_request_error.py
```diff
@@ -0,0 +1,57 @@
+"""Test null response error raising."""
```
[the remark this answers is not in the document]
ElijahSwiftIBM: fixed.

tests/common/test_common_constants.py
ElijahSwiftIBM: fixed.

tests/common/test_common_constants.py
ElijahSwiftIBM: fixed.

pyracf/scripts/setup_precheck.py
ElijahSwiftIBM: sure, fixed.

pyracf/common/downstream_fatal_error.py
ElijahSwiftIBM: fixed.

pyracf/common/downstream_fatal_error.py
ElijahSwiftIBM: good point. This was a relic of old structure.

pyracf/common/security_admin.py
```diff
@@ -79,11 +83,29 @@ def __init__(
             self.__replace_valid_segment_traits(replace_existing_segment_traits)
         if additional_secret_traits is not None:
             self.__add_additional_secret_traits(additional_secret_traits)
+        self.set_running_userid(run_as_userid)
+
+    # ============================================================================
+    # Run as Other User ID
+    # ============================================================================
+    def set_running_userid(self, new_userid: Union[str, None]) -> None:
```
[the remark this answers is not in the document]
ElijahSwiftIBM: fair point. Fixed.

pyracf/common/security_admin.py
ElijahSwiftIBM: I agree it could be superfluous, but I thought that there could be use cases where clearing it explicitly may be preferred, whereas programmatically pulling a userid from an object but having `None` restore defaults felt appropriate.

pyracf/common/downstream_fatal_error.py
```diff
@@ -0,0 +1,66 @@
...
+            )
+        elif (
+            (self.saf_return_code == 8)
+            and (self.racf_return_code == 200)
+            and (self.racf_reason_code == 8)
+        ):
+            self.message += (
+                "\n\nCheck to see if the proper RACF permissions are in place.\n"
+                + "For the `run_as_userid` feature, you must have at least UPDATE"
+                + f" access to `{run_as_userid}.IRRSMO00` in the `SURROGAT` class."
```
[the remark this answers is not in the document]
ElijahSwiftIBM: corrected everywhere I could find it (many places).

pyracf/common/downstream_fatal_error.py
ElijahSwiftIBM: I think this is fine as is. "documented return and reason codes" is appropriate, but I can change it to "documentation's return and reason codes"

pyracf/common/downstream_fatal_error.py
```diff
@@ -0,0 +1,66 @@
+"""Exception to use when IRRSMO00 returns with an error."""
```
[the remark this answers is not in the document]
ElijahSwiftIBM: changed

pyracf/common/security_admin.py
ElijahSwiftIBM: this actually isn't quite the case, I was mistaken. This return code is NOT the SAF return code as that isn't always validly read into our results dictionary. I can make this >=8 or >4 as initially specified. IRRSMO00 documents that the RACF return code is only greater than 4 when the SAF return code is also, but the RACF return code can be LOTS of values in that range.

pyracf/common/security_result.py
lcarcaramo: Testing issues resolved using a mock.

tests/group/group_result_samples/add_group_result_error.json
lcarcaramo: XML format of these errors reverted back to their original format, but are now included with `DownstreamFatalError` instead of `SecurityRequestError`.

tests/group/group_result_samples/add_group_result_error.json
lcarcaramo: These errors have been reverted back to their original format and are now included as part of `DownstreamFatalError` instead of `SecurityRequestError`.

pyracf/common/security_admin.py
lcarcaramo: Ok, if a return code of anything 8 or greater is indicative of a problem with IRRSMO00, then I think a conditional of `>=8` makes sense given that we could assume that any return code lower than 8 would be indicative of a request that processed successfully.

pyracf/common/downstream_fatal_error.py
lcarcaramo: My bad. You are right. `documented` is correct here.

A small suggestion I have however is to move the space from before `the` on line 59 to after `against` on line 58.

lcarcaramo: Actually, I think if you swap `IRRSMO00` and `documented` it works better:

```python
                + "the documented IRRSMO00 return and reason codes for more information "
```

pyracf/common/downstream_fatal_error.py
lcarcaramo: Now that I think about it, it may be even better to just extract these values from the list and make them individual arguments to the constructor:

```python
    def __init__(
        self,
        saf_return_code: int,
        racf_return_code: int,
        racf_reason_code: int,
        request_xml: bytes,
        run_as_userid: Union[str, None] = None,
        result_dictionary: dict = None,
    ) -> None:
```

pyracf/common/logger.py
```diff
@@ -161,7 +161,7 @@ def redact_request_xml(
 
     def redact_result_xml(
         self,
-        xml_string: str,
+        xml_response: Union[str, List[int]],
```
lcarcaramo, line 164: My be better to call this `response` instead of `xml_response` since it may not be XML.

pyracf/common/security_admin.py
```diff
@@ -157,32 +179,59 @@ def _make_request(
...
+            secret_traits=self.__secret_traits,
+        )
         if self._generate_requests_only:
-            request_xml = self.__logger.redact_request_xml(
-                security_request.dump_request_xml(encoding="utf-8"),
-                secret_traits=self.__secret_traits,
-            )
             self.__clear_state(security_request)
             return request_xml
         result_xml = self.__logger.redact_result_xml(
```
lcarcaramo, line 189: Given that `result_xml` may actually not be XML in some situations, it may make more sense to call this variable `result` to avoid confusion about `DownstreamFatalError`.

pyracf/common/security_admin.py
```diff
@@ -157,32 +179,59 @@ def _make_request(
...
+                return_reason_codes=result_xml,
+                request_xml=request_xml,
+                run_as_userid=self.get_running_userid(),
+            )
         if self.__debug:
             # No need to redact anything here since the raw result XML
             # already has secrets redacted when it is built.
             self.__logger.log_xml("Result XML", result_xml)
-        results = SecurityResult(result_xml)
+        results = SecurityResult(result_xml, self.get_running_userid())
```
lcarcaramo, line 211: `results` should be changed to `result` since `results` implies that there are multiple security results.

pyracf/common/security_admin.py
```diff
@@ -79,11 +83,29 @@ def __init__(
...
         if additional_secret_traits is not None:
             self.__add_additional_secret_traits(additional_secret_traits)
+        self.set_running_userid(run_as_userid)
+
+    # ============================================================================
+    # Run as Other User ID
+    # ============================================================================
+    def set_running_userid(self, userid: Union[str, None]) -> None:
+        if userid is None:
+            self.__running_userid = userid
```
lcarcaramo, line 93: change to the following for additional clarity:

```python
self.__running_userid = None
```

pyracf/common/security_admin.py
```diff
@@ -157,32 +179,59 @@ def _make_request(
...
+                    result_dictionary["securityResult"]["returnCode"],
+                    result_dictionary["securityResult"]["reasonCode"],
+                ],
+                request_xml,
+                self.get_running_userid(),
+                result_dictionary,
+            )
         if result_dictionary["securityResult"]["returnCode"] != 0:
             # All non-zero return codes should cause a SecurityRequestError to be raised.
             # Even if a return code of 4 is not indicative of a problem, it it is
```
lcarcaramo, line 237: `it` is repeated. I think this was my mistake.

pyracf/scripts/setup_precheck.py
```diff
@@ -0,0 +1,34 @@
...
+    resource_admin = ResourceAdmin()
+    try:
+        access = resource_admin.get_my_access("IRR.IRRSMO00.PRECHECK", "XFACILIT")
+    except SecurityRequestError:
+        traits_precheck = {"base:universal_access": "None"}
+        result = resource_admin.add(
+            "IRR.IRRSMO00.PRECHECK", "XFACILIT", traits=traits_precheck
+        )
+        print(
+            "IRR.IRRSMO00.PRECHECK is now defined with a `Universal Access` of None.\n"
```
lcarcaramo, line 14:
```python
"IRR.IRRSMO00.PRECHECK is now defined with a 'Universal Access' of 'None'.\n"
```

Note: change (`) to (') for console messages.

pyracf/scripts/setup_precheck.py
```diff
@@ -0,0 +1,34 @@
...
+            + "Contact your security administrator for READ access before using pyRACF.\n"
```
lcarcaramo, line 15:
```python
+ "Contact your security administrator for 'READ' access before using pyRACF.\n"
```

pyracf/scripts/setup_precheck.py
```diff
@@ -0,0 +1,34 @@
...
+            + "Other users of pyRACF will also need to have at least read access.\n"
+            + "You may also need to REFRESH the `XFACILIT` class.\n"
+            + "Review our documentation at https://ambitus.github.io/pyracf/ as well!"
+        )
+        return result
+    if access:
+        print(
+            f"IRR.IRRSMO00.PRECHECK is already defined, and you already have {access} access!\n"
```
lcarcaramo, line 23:
```python
f"'IRR.IRRSMO00.PRECHECK' is already defined, and you already have '{access}' access!\n"
```

pyracf/scripts/setup_precheck.py
```diff
@@ -0,0 +1,34 @@
...
+            + "You are ready to start using pyRACF!\n"
+            + "Please ensure other users of pyRACF also have at least read access.\n"
+            + "Review our documentation at https://ambitus.github.io/pyracf/ as well!"
+        )
+        return True
+    print(
+        "IRR.IRRSMO00.PRECHECK is already defined, but you have no access.\n"
+        + "Contact your security administrator for READ access before using pyRACF.\n"
```
lcarcaramo, line 31:
```python
        "'IRR.IRRSMO00.PRECHECK' is already defined, but you have no access.\n"
        + "Contact your security administrator for 'READ' access before using pyRACF.\n"
```

tests/common/test_downstream_fatal_error.py
```diff
@@ -0,0 +1,57 @@
+"""Test downstream fatal error raising."""
```
lcarcaramo, line 1: Remove `raising`.

tests/common/test_run_as_userid.py
```diff
@@ -0,0 +1,162 @@
...
+        call_racf_mock: Mock,
+    ):
+        userid = "squidwrdtest"
+        precheck_profile_as_squidwrd = (
+            TestCommonConstants.TEST_EXTRACT_RESOURCE_RESULT_PRECHECK_SUCCESS_XML
+        )
+        precheck_profile_as_squidwrd = precheck_profile_as_squidwrd.replace(
+            "<message> 00    ESWIFT          READ              ALTER    NO</message>",
+            "<message> 00    ESWIFT          READ               READ    NO</message>",
+        )
```
lcarcaramo, line 153: Do we need mock data here to test `UserIdError`?

tests/common/test_common_constants.py
```diff
@@ -0,0 +1,143 @@
...
+Sample data for testing Common functions in pyracf.
+"""
+
+from typing import Union
+
+import tests.test_utilities as TestUtilities
+
+
+def get_sample(sample_file: str, function_group: str) -> Union[str, bytes]:
+    return TestUtilities.get_sample(sample_file, function_group)
```
lcarcaramo, line 11: Should we move test sample files specific to these 'common' tests to 'tests/common' so that it is clear what that these test samples are scoped/specific to the 'common' tests?

pyracf/common/downstream_fatal_error.py
ElijahSwiftIBM: Changed to

`python3 
   """
    Raised IRRSMO00 returns with a SAF Return Code of 8,
    indicating that the request could not be processed.
    """
    `

pyracf/common/security_admin.py
ElijahSwiftIBM: Sure, change made.

pyracf/common/security_admin.py
ElijahSwiftIBM: agree, this is appropriate. change made.

pyracf/common/security_admin.py
ElijahSwiftIBM: fair. Change made.

pyracf/common/security_admin.py
ElijahSwiftIBM: fair enough. This is more explicit.

pyracf/common/security_request_error.py
```diff
@@ -1,9 +1,9 @@
-"""Exception to use when data returned by IRRSMO00 indicates that the request failed."""
+"""Exception to use when IRRSMO00 processes a request with RACF warnings or errors."""
 
 
 class SecurityRequestError(Exception):
     """
-    Raised when the return code of a security result returned by IRRSMO00 is NOT equal to 0.
+    Raised when the SAF Return Code of a security result returned by IRRSMO00 is equal to 4.
```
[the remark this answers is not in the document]
ElijahSwiftIBM: really this is only the case for when return code is 4 now that return code of greater than or equal to 8 is `DownstreamFatalError`. We can make it "when the return code is nonzero and a DownstreamFatalError was not raised" or something like that as that is more complete, but it is technically accurate as is.

pyracf/common/security_admin.py
ElijahSwiftIBM: yes, fixed.

pyracf/scripts/setup_precheck.py
ElijahSwiftIBM: fixed.

pyracf/scripts/setup_precheck.py
ElijahSwiftIBM: fixed.

tests/common/test_downstream_fatal_error.py
ElijahSwiftIBM: done.

tests/common/test_run_as_userid.py
ElijahSwiftIBM: No, UserIdError doesn't check RACF for anything, it only looks at basic string formatting. It does "evaluate" the passed string without any mocking necessary.

tests/common/test_common_constants.py
ElijahSwiftIBM: sure. I avoided that in the past as there can be some samples used for both "common" testing and others, but it is probably a good idea. Done.

pyracf/common/security_request_error.py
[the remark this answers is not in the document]
lcarcaramo: Ok, so the idea here is that you either get return code 0, 4, or 8, so it is technically most accurate to say that it gets returned for SAF return code 8?

tests/common/test_run_as_userid.py
lcarcaramo: Let me rephrase the question. It looks like you are setting up mocks here, but if there are not necessary, then there is no point to setting up the mocks, no?

pyracf/common/downstream_fatal_error.py
```diff
@@ -0,0 +1,69 @@
...
+                + "access to 'IRR.IRRSMO00.PRECHECK' in the 'XFACILIT' class."
+            )
+        elif (
+            (self.saf_return_code == 8)
+            and (self.racf_return_code == 200)
+            and (self.racf_reason_code == 8)
+        ):
+            self.message += (
+                "\n\nCheck to see if the proper RACF permissions are in place.\n"
+                + "For the 'run_as_userid' feature, you must have at least UPDATE "
```
lcarcaramo, line 56: `''UPDATE'`

pyracf/scripts/setup_precheck.py
```diff
@@ -0,0 +1,34 @@
...
+    resource_admin = ResourceAdmin()
+    try:
+        access = resource_admin.get_my_access("IRR.IRRSMO00.PRECHECK", "XFACILIT")
+    except SecurityRequestError:
+        traits_precheck = {"base:universal_access": "None"}
+        result = resource_admin.add(
+            "IRR.IRRSMO00.PRECHECK", "XFACILIT", traits=traits_precheck
+        )
+        print(
+            "IRR.IRRSMO00.PRECHECK is now defined with a 'Universal Access' of None.\n"
```
lcarcaramo, line 14: `'NONE'`

pyracf/scripts/setup_precheck.py
```diff
@@ -0,0 +1,34 @@
...
+            + "Contact your security administrator for 'READ' access before using pyRACF.\n"
+            + "Other users of pyRACF will also need to have at least read access.\n"
```
lcarcaramo, line 16: `'READ' access`.

pyracf/scripts/setup_precheck.py
```diff
@@ -0,0 +1,34 @@
...
+            + "You may also need to REFRESH the 'XFACILIT' class.\n"
+            + "Review our documentation at https://ambitus.github.io/pyracf/ as well!"
+        )
+        return result
+    if access:
+        print(
+            f"IRR.IRRSMO00.PRECHECK is already defined, and you already have '{access}' access!\n"
+            + "You are ready to start using pyRACF!\n"
+            + "Please ensure other users of pyRACF also have at least read access.\n"
```
lcarcaramo, line 25: `'READ' access`

pyracf/scripts/setup_precheck.py
```diff
@@ -0,0 +1,34 @@
+from pyracf import ResourceAdmin, SecurityRequestError
+
+
+def setup_precheck():
+    resource_admin = ResourceAdmin()
+    try:
+        access = resource_admin.get_my_access("IRR.IRRSMO00.PRECHECK", "XFACILIT")
```
lcarcaramo, line 7: Should we use `.upper()` on `access` so that `{access}` in the console messages is uppercase for consistency?

pyracf/common/logger.py
ElijahSwiftIBM: sure, changed.

pyracf/common/security_request_error.py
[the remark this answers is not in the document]
ElijahSwiftIBM: yes, but I am okay making it "for any Nonzero return codes that don't raise `DownstreamFatalError`" since that one is also worded "openly"

tests/common/test_common_constants.py
ElijahSwiftIBM: I can make it so that no data is duplicated, but it may make the common constants file a bit messy indicating it pulls test data from different places.

pyracf/common/downstream_fatal_error.py
```diff
@@ -0,0 +1,69 @@
...
+            )
+            self.result = result_dictionary
+        elif (
+            (self.saf_return_code == 8)
+            and (self.racf_return_code == 200)
+            and (self.racf_reason_code == 16)
+        ):
+            self.message += (
+                "\n\nCheck to see if the proper RACF permissions are in place.\n"
+                + "For 'set' or 'alter' functions, you must have at least READ "
```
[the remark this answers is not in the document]
ElijahSwiftIBM: fixed.

pyracf/common/security_admin.py
```diff
@@ -157,35 +176,62 @@ def _make_request(
...
+                racf_reason_code=result_dictionary["securityResult"]["reasonCode"],
+                request_xml=request_xml,
+                run_as_userid=self.get_running_userid(),
+                result_dictionary=result_dictionary,
             )
-        result_dictionary = results.get_result_dictionary()
         if result_dictionary["securityResult"]["returnCode"] != 0:
             # All non-zero return codes should cause a SecurityRequestError to be raised.
-            # Even if a return code of 4 is not indicative of a problem, it it is
+            # Even if a return code of 4 is not indicative of a problem, it is
```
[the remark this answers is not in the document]
ElijahSwiftIBM: the only listed return code 4 situations in IRRSMO00's doc is for both SAF and RACF RC 4. Theoretically this is true for either, and I think that specifying SAF could lead users to think RACF RC4 IS always indicative of a problem, whereas the more general statement covers both cases. I did tweak the wording and spacing a bit anyway.

pyracf/scripts/setup_precheck.py
ElijahSwiftIBM: fixed.

pyracf/scripts/setup_precheck.py
ElijahSwiftIBM: fixed.

pyracf/scripts/setup_precheck.py
ElijahSwiftIBM: fixed.

pyracf/scripts/setup_precheck.py
ElijahSwiftIBM: was just thinking this, it seems appropriate. fixed.

tests/common/test_common_constants.py
lcarcaramo: There is clearly a tradeoff here, but I do think that a little bit of test data duplication here is okay since it makes our test code easier to understand since it is clearer where test data is coming from.

However, a possible solution to this problem to this problem is to maybe use symbolic links maybe? I think git can handle them, but I don't know if they would work properly on Windows, so at the end of the day it might just be better to just have a little bit of data duplication. Let me know what you think.

pyracf/common/security_admin.py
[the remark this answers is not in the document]
lcarcaramo: So in the comment should we say something about the SAF return code always matching the RACF return code?

pyracf/common/security_request_error.py
```diff
@@ -21,6 +21,7 @@ def __str__(self) -> str:
     def contains_error_message(
         self, security_definition_tag: str, error_message_id: str
     ):
+        """Checks to see if specific error message id appears in the security request error"""
```
lcarcaramo, line 24: Add `.` to the end.

pyracf/scripts/setup_precheck.py
```diff
@@ -0,0 +1,35 @@
...
+        print(
+            "IRR.IRRSMO00.PRECHECK is already defined, "
+            + f"and you already have '{access.upper()}' access!\n"
+            + "You are ready to start using pyRACF!\n"
+            + "Please ensure other users of pyRACF also have at least 'READ' access.\n"
+            + "Review our documentation at https://ambitus.github.io/pyracf/ as well!"
+        )
+        return True
+    print(
+        "IRR.IRRSMO00.PRECHECK is already defined, but you have no access.\n"
```
lcarcaramo, line 31: `'IRR.IRRSMO00.PRECHECK'`

tests/common/test_downstream_fatal_error.py
```diff
@@ -0,0 +1,57 @@
...
+__init__
+
+
+@patch("pyracf.common.irrsmo00.IRRSMO00.call_racf")
+class TestDownstreamFatalError(unittest.TestCase):
+    maxDiff = None
+    IRRSMO00.__init__ = Mock(return_value=None)
+    user_admin = UserAdmin()
+
+    def test_security_request_error_thrown_on_precheck_error(
```
lcarcaramo, line 22: `downstream_fatal_error` not `security_request_error`

tests/test_utilities.py
```diff
@@ -57,7 +57,7 @@ def get_sample(sample_file: str, function_group: str) -> Union[str, bytes]:
     result_samples = f"{function_group}/{function_group}_result_samples"
     request_samples = f"{function_group}/{function_group}_request_samples"
     log_samples = f"{function_group}/{function_group}_log_samples"
-    if f"{function_group}_result" in sample_file:
+    if "_result" in sample_file:
```
lcarcaramo, line 60: Is this change necessary now that we are not sharing test data between function groups?

pyracf/common/security_request_error.py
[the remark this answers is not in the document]
ElijahSwiftIBM: Changed to:

` python
   """
    Raised when pyRACF does not rais a DownstreamFatalError, and the SAF Return Code 
    of a security result returned by IRRSMO00 is non-zero.
    """
`

tests/common/test_common_constants.py
ElijahSwiftIBM: Sounds good, we can definitely put a streamlining of test data/cases on the pipeline for the future.

pyracf/common/security_admin.py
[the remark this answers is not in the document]
ElijahSwiftIBM: I mean in our dictionary it is just listed as a "returnCode" at this stage, and I think that adding the distinction in the comments makes that structure harder to follow.

pyracf/common/security_request_error.py
ElijahSwiftIBM: done.

pyracf/scripts/setup_precheck.py
```diff
@@ -0,0 +1,35 @@
...
+    resource_admin = ResourceAdmin()
+    try:
+        access = resource_admin.get_my_access("IRR.IRRSMO00.PRECHECK", "XFACILIT")
+    except SecurityRequestError:
+        traits_precheck = {"base:universal_access": "None"}
+        result = resource_admin.add(
+            "IRR.IRRSMO00.PRECHECK", "XFACILIT", traits=traits_precheck
+        )
+        print(
+            "IRR.IRRSMO00.PRECHECK is now defined with a 'Universal Access' of 'NONE'.\n"
```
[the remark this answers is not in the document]
ElijahSwiftIBM: change made in all relevant locations.

pyracf/scripts/setup_precheck.py
```diff
@@ -0,0 +1,35 @@
...
+            + "Contact your security administrator for 'READ' access before using pyRACF.\n"
+            + "Other users of pyRACF will also need to have at least 'READ' access.\n"
+            + "You may also need to REFRESH the 'XFACILIT' class.\n"
```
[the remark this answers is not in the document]
ElijahSwiftIBM: dhanged.

tests/common/test_downstream_fatal_error.py
ElijahSwiftIBM: fixed in all locations here.

pyracf/scripts/setup_precheck.py
```diff
@@ -0,0 +1,35 @@
...
+            + "Review our documentation at https://ambitus.github.io/pyracf/ as well!"
+        )
+        return result
+    if access:
+        print(
+            "IRR.IRRSMO00.PRECHECK is already defined, "
```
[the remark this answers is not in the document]
ElijahSwiftIBM: change made in all relevant locations.

pyracf/scripts/setup_precheck.py
ElijahSwiftIBM: change made in all relevant locations.

tests/test_utilities.py
ElijahSwiftIBM: technically not. Names of "common test files could be changed to all have "common_result" in their name when necessary, but I think this makes it harder to distinguish what sort of admin activity is occurring in the common result files, so i thought that keeping the "user_result" or "resource_result" in the names worked best, but with that in place, without the above change, the code could not find them.

tests/common/test_setup_precheck.py
```diff
@@ -0,0 +1,157 @@
...
+import tests.common.test_common_constants as TestCommonConstants
+from pyracf import setup_precheck
+from pyracf.common.irrsmo00 import IRRSMO00
+
+# Resolves F401
+__init__
+
+
+@patch("pyracf.common.irrsmo00.IRRSMO00.call_racf")
+class TestSetupPrecheck(unittest.TestCase):
```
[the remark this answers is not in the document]
ElijahSwiftIBM: Theoretically yes, I understand where this makes sense, and would like to do it. Functionally, I cannot think of a case where this would be true. Our `SecurityRequestError` should not be raised by this script as it is currently defined. The Profile name meets the criteria for the XFACILIT class, and XFACILIT is a default class included with RACF. This also doesn't meet any known criteria for `DownstreamFatalError` as it is not running as a userid or performing any "alter" options that would require precheck. It is possible that it is capable of failure, but its surfacing of the error message is just a matter of python bringing that up as it has no special path of its own to raise such an exception, so I feel like this is functionally tested by our tests of `add_resource_profile`. I am open to further discussion though.

pyracf/common/security_request_error.py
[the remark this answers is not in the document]
lcarcaramo: Add an `e` to raise.

pyracf/common/security_admin.py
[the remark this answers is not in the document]
lcarcaramo: Good point. We can leave it as referring to `returnCode` in the result dictionary since talking about SAF and RACF might make things more confusing. I agree, keeping it simple and just referring to `returnCode` is better in that sense.

tests/common/test_setup_precheck.py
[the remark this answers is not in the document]
lcarcaramo: That is a good point. We know that add resource can raise an exception and that has already technically been tested. However, I think it would be good to have a test just to demonstrate the exception raised from failing to add the resource gets manifested correctly.

pyracf/common/security_request_error.py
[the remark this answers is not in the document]
ElijahSwiftIBM: good catch.

tests/test_utilities.py
lcarcaramo: Ok, I see, the problem is that the sample files in common don't start with `common`. In that case, I think this change makes sense.

tests/common/test_setup_precheck.py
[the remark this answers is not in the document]
lcarcaramo: I think that should be sufficient. I think the key here is to demonstrate that if a `SecurityRequestError` is raised during the add, it gets manifested properly.

My only suggestion here is to change the name of the test function to `test_setup_precheck_throws_error_when_add_resource_fails`

pyracf/common/security_admin.py
lcarcaramo: This conversation is still unresolved.

tests/common/test_setup_precheck.py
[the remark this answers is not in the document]
ElijahSwiftIBM: changed.

pyracf/common/security_admin.py
ElijahSwiftIBM: Sorry, this got lumped in and hidden with a bunch of conversations that were resolved on my view. Thank you for bringing my attention back to it.

I can easily change this comment to "greater than or equal to 8" to reflect the code more directly. These are all indicative of an issue with IRRSMO00 that would stop a command image from being generated.

The comment now reads:

```python
            # All return codes greater than or equal to 8 are indicative of an issue
            # with IRRSMO00 that would stop RACF from generating a command image.
            # The user should interogate the result dictionary attached to the
            # SecurityRequestError as well as the return and reason codes to resolve
            # the problem.
```

tests/common/test_common_constants.py
```diff
@@ -0,0 +1,148 @@
...
+TEST_ALTER_USER_RESULT_SUCCESS_XML = get_sample("alter_user_result_success.xml")
+TEST_EXTRACT_RESOURCE_PRECHECK_AS_SQUIDWRD_LOG = get_sample(
+    "extract_resource_precheck_as_squidwrd.log"
+)
+
+# ============================================================================
+# Downstream Fatal Error
+# ============================================================================
+
+TEST_DOWNSTREAM_FATAL_PRECHECK_TEXT = (
```
lcarcaramo, line 125: `TEST_DOWNSTREAME_FATAL_ERROR`

tests/group/test_group_result_parser.py
```diff
@@ -62,6 +64,37 @@ def test_group_admin_throws_error_on_add_existing_group(
             + f"'{group_name}' already exists as a '{self.group_admin._profile_type}' profile.",
         )
 
+    def test_group_admin_add_surfaces_error_from_preliminary_extract(
```
lcarcaramo, line 67: What is this testing for?

tests/group/test_group_result_parser.py
```diff
@@ -62,6 +64,37 @@ def test_group_admin_throws_error_on_add_existing_group(
...
+        call_racf_mock.side_effect = [
+            extract_group_error_xml,
+            TestGroupConstants.TEST_ADD_GROUP_RESULT_SUCCESS_XML,
+        ]
+        with self.assertRaises(SecurityRequestError) as exception:
+            self.group_admin.add(
+                group_name, traits=TestGroupConstants.TEST_ADD_GROUP_REQUEST_TRAITS
+            )
+        extract_group_error_json = copy.deepcopy(
+            TestGroupConstants.TEST_EXTRACT_GROUP_RESULT_BASE_ONLY_ERROR_DICTIONARY
```
lcarcaramo, line 88: I don't think you need `copy.deepcopy` here. I am pretty sure that you can just use `dict(my_dictionary)` to create a deep copy.

ElijahSwiftIBM: Unfortunately, this did not work. I did try that first. The issue is because the dictionary contains other dictionaries `dict(my_dictionary)` was making a shallow copy, which meant that the modification made to `extract_group_error_json` was carried to other values in TEST_EXTRACT_GROUP_RESULT_BASE_ONLY_ERROR_DICTIONARY.

tests/group/test_group_result_parser.py
ElijahSwiftIBM: This testing enters the path in the coverage report that was previously missed, where EXTRACT returns an error that DOES NOT contain the error message indicating the group already exists. In the event of this, group admin (and others like it) surface the initial error directly, so this test covers that codepath.

tests/common/test_common_constants.py
ElijahSwiftIBM: This is more complete, now this is TEST_DOWNSTREAM_FATAL_ERROR_PRECHECK_TEXT and the same change has been made for similar constants.

tests/group/test_group_result_parser.py
lcarcaramo: Okay, this covers the code coverage issue. Got it.

lcarcaramo: I still think this is a good test to have, but is there a real world situation where this would happen? I see that you just substitute a fake error message to cover that code path in this example.

pyracf/scripts/setup_precheck.py
```diff
@@ -0,0 +1,35 @@
...
+
+def setup_precheck():
+    resource_admin = ResourceAdmin()
+    try:
+        access = resource_admin.get_my_access("IRR.IRRSMO00.PRECHECK", "XFACILIT")
+    except SecurityRequestError:
+        traits_precheck = {"base:universal_access": "None"}
+        result = resource_admin.add(
+            "IRR.IRRSMO00.PRECHECK", "XFACILIT", traits=traits_precheck
+        )
```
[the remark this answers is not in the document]
ElijahSwiftIBM: If the XFACILIT class is RACLISTED then yes, otherwise no. I will change the script to do that where necessary. I'll also add unit testing that accounts for authority that the user may or may not have to actually do that.

tests/group/test_group_result_parser.py
ElijahSwiftIBM: I would suspect this may be the case if the user doesnt have the authority to issue the "extract" command (LISTGRP for Group), but I would also imagine that there are other cases. I guess since the purpose was to test any error message that wasn't the specific one we were looking for, I was aiming to just substitute in anything, but I can look for the appropriate error messages as well.

pyracf/scripts/setup_precheck.py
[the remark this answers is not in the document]
ElijahSwiftIBM: Oh I did actually say "You may also need to refresh the XFACILIT class" in the message printed out if the class gets added. I could make it do this programmatically, but there could be some hiccups with user authority that it could trip over, and I wanted to keep this simple since anyone can theoretically run this. I can change it though, if you think that would be best.

tests/group/test_group_result_parser.py
ElijahSwiftIBM: This has been added in the latest commit.

pyracf/scripts/setup_precheck.py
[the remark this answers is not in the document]
lcarcaramo: Right, I see that in the message. Given that the setropts refresh could cause issues in some cases, I think this is probably good enough, especially because this nuance is mentioned in the message.

pyracf/common/downstream_fatal_error.py
```diff
@@ -0,0 +1,69 @@
+"""Exception to use when IRRSMO00 is unable to process a request."""
+from typing import Union
+
+
+class DownstreamFatalError(Exception):
+    """
+    Raised IRRSMO00 returns with a SAF Return Code of 8,
```
lcarcaramo, line 7: `Raised when`

pyracf/common/security_admin.py
```diff
@@ -79,11 +83,28 @@ def __init__(
...
+    # ============================================================================
+    # Run as Other User ID
+    # ============================================================================
+    def set_running_userid(self, userid: Union[str, None]) -> None:
+        if userid is None:
+            self.__running_userid = None
+            return
+        if not isinstance(userid, str) or len(userid) > 8 or userid == "":
+            raise UserIdError(userid)
+        self.__running_userid = userid.upper()
```
lcarcaramo, line 97: Do we want to convert the userid to lowercase here instead of in `get_running_userid()`?

If we do that, then `get_running_userid()` is just this:
```python
    def get_running_userid(self) -> Union[str, None]:
        return self.__running_userid
```
Also need to specify that the return value is `Union[str,None]`.

ElijahSwiftIBM: Change made to Union[str, None] in type hints. Unfortunately, while this CAN be done, I think this is the best case scenario. After testing, it seems like we would NEED to ensure the "__running_userid" value is uppercase before passing it to RACF, so checks to drive the string to upper and lowercase will have to go somewhere, and validity checks to ensure that the values ARE strings that can be driven to upper or lower case are also necessary. I would rather have all of this done ONLY when the values are changed or grabbed rather than for every request made to RACF, but I am open to further discussion.

lcarcaramo: That is fine. I think we can revisit this later since this was pretty minor. Let's get this release out.

lcarcaramo (approved):

Issue #61, opened 2024-01-08
Bug Report: SecurityRequestError Doc references out of date


ElijahSwiftIBM:
**Describe the bug**
Documentation error in 1.0b3. Numerous functions document SecurityRequestError as 

`Raises `SecurityRequestError` when the **Return Code** of a **Security Result** returned by IRRSMO00 is **NOT** equal to `0``

This needs to be updated and all functions capable of raising SecurityRequestError in this way can also raise DownstreamFatalError as well.

**To Reproduce**
N/A

**Expected behavior**
Change documentation to

`Raises `SecurityRequestError` when the **Return Code** of a **Security Result** returned by IRRSMO00 is equal to `4``

also include comparable documentation for DownstreamFatalError

**Console Output**
N/A

lcarcaramo:
Resolved by #62

PR #52, opened 2023-11-21
Add Documentation for New Audit Rules Functions

ElijahSwiftIBM:
-Add Documentation for New Audit Rules Functions
-Document now stable Audit Rules traits in base segment

### :bulb: Issue Reference

**Issue:**  #40 

### :computer: What does this address?

As part of PR #51 traits were stabilized and new functions added. This PR covers the documentation.

### :pager: Implementation Details

N/A

### :clipboard: Is there a test case?

N/A

PR #62, opened 2024-01-08
Update function error references

ElijahSwiftIBM:
-Change SecurityRequestError references within documented functions to correct misleading RC information -Add references to DownstreamFatalError in all functions capable of returning SecurityRequestError

### :bulb: Issue Reference

**Issue:**  #61 

Documentation error in 1.0b3. Numerous functions document SecurityRequestError as

Raises SecurityRequestErrorwhen the **Return Code** of a **Security Result** returned by IRRSMO00 is **NOT** equal to0``

This needs to be updated and all functions capable of raising SecurityRequestError in this way can also raise DownstreamFatalError as well.

N/A

N/A


resource/abstractions/auditing_rules.md
```diff
@@ -0,0 +1,474 @@
+---
```
lcarcaramo, line 1: The Auditing rules documentation shouldn't be included in this PR. These updates are part of #52

ElijahSwiftIBM: Correct, not sure exactly how this happened but it's been removed.

user/abstractions/operations.md
```diff
@@ -29,7 +29,9 @@ Check if a user has **Operations** authority.
 
 #### ❌ Raises
 * `SecurityRequestError`<br>
-  Raises `SecurityRequestError` when the **Return Code** of a **Security Result** returned by IRRSMO00 is **NOT** equal to `0`.
+  Raises `SecurityRequestError` when the **Return Code** of a **Security Result** returned by IRRSMO00 is equal to `4`.
+* `DownstreamFatalError`<br>
+  Raises `DownstreamFatalError` when the **Return Code** of a **Security Result** returned by IRRSMO00 is greater than `4`.
 
 #### 💻 Example
```
lcarcaramo, line 37: The exceptions for `take_away_operations_authority()` still needs to be updated.

access/base/delete.md
```diff
@@ -44,7 +44,9 @@ Delete a specified **permission**
 
 #### ❌ Raises
 * `SecurityRequestError`<br>
-  Raises `SecurityRequestError` when the **Return Code** of a **Security Result** returned by IRRSMO00 is **NOT** equal to `0`.
+  Raises `SecurityRequestError` when the **Return Code** of a **Security Result** returned by IRRSMO00 is equal to `4`.
```
lcarcaramo, line 47: I think we can get rid of the `returned by IRRSMO00 part` to reduce noise. I think the following gets the message across:

Also, for clarity, I think we should be referring to the `SAF Return Code`.

```markdown
Raises `SecurityRequestError` when the **SAF Return Code** of a **Security Result** is equal to `4`.
```

user/abstractions/operations.md
ElijahSwiftIBM: Not sure how `operations.md` got missed in my initial crawl but there were two places that needed a touch up. Both addressed.

access/base/delete.md
lcarcaramo: This issue has been resolved.

PR #64, opened 2024-01-19
Gh pages feature/add optional expired

ElijahSwiftIBM:
### :bulb: Issue Reference

**Issue:**  #49 

Doc changes to accompany #55 

N/A

N/A

### DIFF
--- a/user/abstractions/passphrase.md
+++ b/user/abstractions/passphrase.md
@@ -14,7 +14,7 @@ User Administration functions for modifying a user's passphrase.
 
 <br>
 ```python
-def set_passphrase(self, userid: str, passphrase: str) -> Union[dict, bytes]:
+def set_passphrase(self, userid: str, passphrase: str, expired: bool = False) -> Union[dict, bytes]:
 ```
 
 #### 📄 Description
@@ -40,6 +40,9 @@ Set a user's **passphrase**.
 
 * `passphrase`<br>
   The **passphrase** to set for the specified user or `False` to delete the current value.
+
+* `expired`<br>
+  A bool representing whether or not the **passphrase** should be set as **expired** and require changing on next logon.
 
 #### 📤 Returns
 
--- a/user/abstractions/passphrase.md
+++ b/user/abstractions/passphrase.md
@@ -14,7 +14,9 @@ User Administration functions for modifying a user's passphrase.
 
 <br>
 ```python
-def set_passphrase(self, userid: str, passphrase: str, expired: Union[bool, None] = None) -> Union[dict, bytes]:
+def set_passphrase(
+    self,  userid: str, passphrase: str, expired: Union[bool, None] = None
+) -> Union[dict, bytes]:
 ```
 
 #### 📄 Description
--- a/user/abstractions/passphrase.md
+++ b/user/abstractions/passphrase.md
@@ -14,7 +14,7 @@ User Administration functions for modifying a user's passphrase.
 
 <br>
 ```python
-def set_passphrase(self, userid: str, passphrase: str, expired: bool = False) -> Union[dict, bytes]:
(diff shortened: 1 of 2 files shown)


user/abstractions/passphrase.md
```diff
@@ -41,6 +41,9 @@ Set a user's **passphrase**.
 * `passphrase`<br>
   The **passphrase** to set for the specified user or `False` to delete the current value.
 
+* `expired`<br>
+  A bool representing whether or not the **passphrase** should be set as **expired** and require changing on next logon.
```
lcarcaramo, line 45:
```markdown
* `expired`<br>
  A boolean toggle that is used to determine whether or not the user's **passphrase** should be set as **expired**,  meaning that the user will be required to change their **passphrase** on their next logon. If this argument is left unset, the user's **passphrase** will be set as **expired** by default.
```

Do the same for password as well.

I am pretty sure that the second sentence is correct, but if not, then it might be good to say that passwords/passphrase set this way usually default to expired and then just briefly explain the edge case.

ElijahSwiftIBM: With the current code, the explanation is arguably correct, but I don't love it because it implies that if you don't supply `expired`, pyracf basically sets it to True which is not how the code is written (anymore) I changed it to the following.


```
* `expired`<br>
  A boolean toggle that is used to determine whether or not the user's **passphrase** should be set as **expired**,  meaning that the user will be required to change their **passphrase** on next logon. Without this argument specified, RACF will maek the user's **passphrase** as **expired** by default.
```

lcarcaramo: So, given that we are not sure what the default behavior would be for security managers and because we don't want to imply that pyRACF is setting the `expired` flag, we specifically say that RACF will set the password/passphrase as expired by default.

Given, that I agree with this change. My only suggestion is a couple changes relating to prose and typo, and then I think we are good with this PR.

```markdown
* `expired`<br>
  A boolean toggle that is used to determine whether or not the user's **passphrase** should be set as **expired**,  meaning that the user will be required to change their **passphrase** on next logon. If this argument is not provided, RACF will set the user's **passphrase** as **expired** by default.
```

ElijahSwiftIBM: Makes sense. This has been changed in both files now.

user/abstractions/password.md
```diff
@@ -41,6 +41,9 @@ Set a user's **password**.
 * `password`<br>
   The **password** to set for the specified user or `False` to delete the current value.
 
+* `expired`<br>
+  A boolean toggle that is used to determine whether or not the user's **password** should be set as **expired**,  meaning that the user will be required to change their **password** on next logon. If this argument is not provided, RACF will set the user's **passpword** as **expired** by default.
```
lcarcaramo, line 45: typo
`password` not `passpword`

user/abstractions/passphrase.md
```diff
@@ -14,7 +14,7 @@ User Administration functions for modifying a user's passphrase.
 
 <br>
 ```python
-def set_passphrase(self, userid: str, passphrase: str) -> Union[dict, bytes]:
+def set_passphrase(self, userid: str, passphrase: str, expired: Union[bool, None] = None) -> Union[dict, bytes]:
```
lcarcaramo, line 17:
```python
def set_passphrase(
    self,  userid: str, passphrase: str, expired: Union[bool, None] = None
) -> Union[dict, bytes]:
```

Do the same for `set_password`.

user/abstractions/password.md
ElijahSwiftIBM: that is an embarrassing typo. Thank you for catching it.

lcarcaramo (approved):

Issue #60, opened 2024-01-05
Feature Request: RAS Enhancements


lcarcaramo:
**Is your feature request related to a problem? Please describe.**
* If a response returned by IRRSMO00 is either ~unable to be decoded using the `cp1047` codec~ *(encoding problems will be captured by XML parsing failures)* or the returned XML is unable to be parsed, the user may not have enough information for problem determination.
* ~There is no way to audit audit RACF commands that get run through IRRSMO00.~ *(Revisit)*

**Describe the solution you'd like**
* If a response returned by IRRSMO00 is either ~unable to be decoded using the `cp1047` codec~ *(encoding problems will be captured by XML parsing failures)* or the returned XML is unable to be parsed, a raw dump of the response should be created.
* The `~/.pyracf/dumps` folder is where raw dumps should be stored.
* ~`~/.pyracf/audit.log` is where executed RACF commands will be written to.~ *(Revisit)*
* ~`~/.pyracf/pyracf.conf` will be used to enable or disable audit logging.~ *(Revisit)*

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

lcarcaramo:
Implemented in #70

PR #70, opened 2024-02-06
RAS Update

lcarcaramo:
### :bulb: Issue Reference

**Issue:**  https://github.com/ambitus/pyracf/issues/60

* When the result XML returned by IRRSMO00 cannot be parsed, a dump of the raw result XML returned by IRRSMO00 should be created to aid in problem determination.
* All post processing of the result returned by IRRSMO00 should be done on the Python side. The C code should return an exact copy of what IRRSMO00 returned back to Python.
* Enable testing of `IRRSMO00.call_racf()`
* Add testing to make sure all class attributes work with all function groups.
* Update dependency versions.
* Add `dump_mode` toggle to force the creation of dumps of the raw result XML returned by IRRSMO00 on all requests (failure and success).
* Add `irrsmo00_result_buffer_size` to enable the size of the IRRSMO00 result buffer to be customized.
* IRRSMO00 result buffer size defaults to 16KB.
* IRRSMO00 result buffer size cannot be less than `10000` and should not be larger than `100000000` since it can lead to a `SIGKILL` signal being raised, which will result in the Python process that pyRACF runs under being kill.
* Fixed bug in definition of `_valid_segment_traits` dictionary for `DataSetAdmin`. Updating `_valid_segment_traits` after calling the parent constructor breaks "Replace Existing Segment Traits" and "Update Existing Segment Traits" functionally. Moving this code to before calling the parent constructor fixes the problem.
* Log warning messages for functionality that should only be used for development and debugging
* Log warning messages for functionality that is experimental.
* Added folders for exceptions and utilities for better organization.
* Add `Dumper` for creation of dump files.
* Hex  Dumps get written to the console when `debug` is enabled and a dump of the raw IRRSMO00 result XML is created.
* Hex Dumps written to the console are redacted.
* raw dumps written to the file system are NOT redacted. These files are protected by being created with `600` permissions and IRRSMO00 already redacts passwords and passphrases which are among the most sensitive information that could be included in a security result. So, given the dump files are protected and sensitive information is more or less already redacted by IRRSMO00, it makes seems to dump the raw result XML from IRRSMO00 with no post processing at all to ensure that during problem determination, an exact copy of what IRRSMO00 returned is available to better determine if there is a problem with IRRSMO00 or if there is a problem pyRACF.

* "Null byte fix logic" moved from the C side to the Python side.
* Use `y#` in `Py_BuildValue()` to return an unmodified copy of the raw result XML from IRRSMO00.
* Dump files are created at `~/.pyracf/dumps` with the naming convention `pyracf.<timestamp>.<md5>.dump`.
* All folders under `~/.pyracf` (including the `.pyracf` folder itself) are created with `700` permissions and dump files are created with `600` permissions.
* If any folders under `~/.pyracf` (including the `.pyracf` folder itself) have any permissions other than `700`, they will be dynamically updated during dump processing. This handles scenarios where the user sets incorrect/insecure permissions on these directories and when `umask` changes the permissions on folder creation.
* Mocking `IRRSMO00.__init__()` is no longer necessary since updating the C code to function as a Python Extension and it also breaks new test cases introduced here. All Mocks of `IRRSMO00.__init__()` have been removed.

Run test cases and try out new functionality manually.


pyracf/common/irrsmo00.py
```diff
@@ -59,52 +59,34 @@ def call_racf(
...
-            len(running_userid),
+        result = call_irrsmo00(
+            request_xml=request_xml,
+            request_xml_length=len(request_xml),
+            result_buffer_size=self.__result_buffer_size,
+            irrsmo00_options=irrsmo00_options,
+            running_userid=running_userid,
+            running_userid_length=len(running_userid),
         )
         # Preserve raw binary respone just in case we need to create a dump.
```
ElijahSwiftIBM, line 70: Typo: `respone` should be `response`

ElijahSwiftIBM (commented):
All looks good minus 1 minor typo. I do like the additional validation you added for buffer size. Good idea, and the test for it looks solid.

pyracf/common/irrsmo00.py
lcarcaramo: This typo has been fixed. Actually, I have made the naming convention consistent by referring to this as the "raw result XML" instead of "raw binary response" for better clarity and consistency. I must have missed this one because of the typo. This should be good now.

ElijahSwiftIBM (approved):
Looks great!

PR #75, opened 2024-02-13
Release 1.0b5

lcarcaramo:
### :bulb: Issue Reference

https://github.com/ambitus/pyracf/milestone/6

Refer to GitHub Milestone.

Refer to GitHub Milestone.

Refer to GitHub Milestone.


ElijahSwiftIBM (commented):
I think this overall looks REALLY good, but I think I want to back away from an earlier decision I made. in RunAsUserid and IndividualAccess we have text like the following:

`...the caller must have at least UPDATE access to the userid.IRRSMO00 resource in the SURROGAT class, where userid represents the specific userid you wish to execute commands as. Further information can be found outlined in [Our Dependencies Note].`

I think that the reference to the Dependencies Note is not a good idea. We should link directly to the same documentation. The Dependencies Note makes sense for PRECHECK authority references like SetupPrecheck and DownstreamFatalError, but the Note doesn't mention the SURROGAT dependencies at all, so for more information you would REALLY have to go all the way to IRRSMO00 documentation. If you agree, I can make those changes and we can test and approve/merge.

lcarcaramo:
I agree. I will close this pull request to allow for the necessary changes to be made to the "Run as Userid" and "Individual Access" documentation.

Issue #71, opened 2024-02-06
Feature Request: pyRACF should Leverage IRRSMO00's ability to compensate for small buffers


ElijahSwiftIBM:
**Is your feature request related to a problem? Please describe.**
pyRACF is about to start using a dynamic buffer size to interact with IRRSMO00 which means it is possible for a user or application to specify a buffer too small for command output. This will result in a `DownstreamFatalError` surfacing IRRSMO00 return codes of `8/4000/Len`, but the partial data would be discarded and this would require the application or user to re-drive the activity with a larger buffer which is undesirable.

**Describe the solution you'd like**
Note: It is possible for IRRSMO00 to process a part of the input document, and then find there is not enough space in the result buffer to return the results. In this situation, the results for that portion of the input are lost, and that part of the request is repeated on the next call to IRRSMO00. This may result in unexpected output, such as a 'not found' error if the work in question was a delete operation. This can only happen If the result is too small.

https://www.ibm.com/docs/en/zos/3.1.0?topic=rsismo-usage-notes

**Describe alternatives you've considered**
We could try to implement this at the C level, but it would complicate the simple nature of its existing structure, and it would also require a lot of dynamic memory management as there would need to be a lot of "simultaneous" storage held to build the buffer into a large response before returning to the python layer.

We could try to implement this at the python layer, but I am uncertain all the necessary memory objects will exist and still be accessible when a call comes in to re-drive the c code from python with the added handle pointer.

**Additional context**
N/A

lcarcaramo:
Since RACFu will be the backend for pyRACF and RACFu already does this, this is no longer relevant for pyRACF.

PR #79, opened 2024-02-20
Feature/redrive on buffer too small error

ElijahSwiftIBM:
### :bulb: Issue Reference

**Issue:**  #71 

Now that users can specify buffer sizes as a part of the class constructor, we need to dynamically respond to the buffer size being too small to hold the response xml.

At the C code level, if the failing request is detected, a new buffer is allocated large enough for the remaining response and the request is re-driven. If the buffer would need to be larger than the maximum tolerable size for our constructor, we fail and surface the error as a DownstreamFatalError with IRRSMO00's return and reason codes.

Since this re-drive is done at the c level, there are no applicable unit tests available. A function test could be added, but we would need confirmation that the environment is appropriate for the test (a command and buffer size must be chosen that we KNOW would fail prior to this feature which may differ dependent on the environment and authority bestowed to pyRACF).

On our test system, I ran two tests. One with setropts_admin.list_racf_options with a buffer specified at 10000 bytes (the minimum) which failed before the added code and now succeeds, and one with resource_admin.extract("*","FACILITY) with the default 16KB buffer which failed and now succeeds.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@
 
 [tool.poetry]
     name="pyracf"
-    version="1.0b5"
+    version="1.0b6"
     description="Python interface to RACF using IRRSMO00 RACF Callable Service."
     license = "Apache-2.0"
     authors = [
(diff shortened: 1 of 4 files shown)


ElijahSwiftIBM:
Closing this PR as we are re-assessing these features. This will either get moved to python layer or involve new unit testing in C layer.

Issue #68, opened 2024-01-31
Feature Request: Improve Additional Secrets Redaction


ElijahSwiftIBM:
**Is your feature request related to a problem? Please describe.**
Additional Secrets Redaction is currently marked experimental because messages returned by IRRSMO00 from RACF can contain the redacted content. Additionally the regex implementation of the redaction is not exhaustive and seems that it could be error prone if the weaknesses of it were deliberately exploited.

**Describe the solution you'd like**
Error messages containing redacted content should also be redacted.
Redaction should be employed in a way to minimize false positives and guarantee redaction of the target pattern.

**Describe alternatives you've considered**
No alternatives as there is no defined implementation for this

**Additional context**
N/A

PR #80, opened 2024-02-20
Feature/additional secrets improvements

ElijahSwiftIBM:
### :bulb: Issue Reference

**Issue:**  #68 

Additional Secrets Redaction was marked experimental due to 2 key weaknesses. 1) not redacting secret material in messages in the security response object and 2) the method of secret redaction opened the possibility of incomplete redaction of "additional" secrets.

Additional Secrets redaction now uses more streamlined regex pattern matching to exhaustively redact secrets in the command images. Additional features were added to redact messages by their RACF tag to ensure that no "false positives" would give away redacted fields.

Additional test cases were added/existing test cases were changed to test the new behavior of redacting more complex and varied fields as well as messages.

### DIFF
--- a/tests/common/test_additional_secrets_redaction.py
+++ b/tests/common/test_additional_secrets_redaction.py
@@ -30,6 +30,7 @@ __init__
 
 class TestAdditionalSecretsResultRedaction(unittest.TestCase):
     maxDiff = None
+    ansi_escape = re.compile(r"\x1B(?:[@-Z\\-_]|\[[0-?]*[ -/]*[@-~])")
 
     # ============================================================================
     # User Administration
@@ -119,6 +120,61 @@ class TestAdditionalSecretsResultRedaction(unittest.TestCase):
             str(result),
         )
 
+    @patch("pyracf.common.irrsmo00.IRRSMO00.call_racf")
+    def test_alter_user_request_debug_log_additional_secret_added_get_redacted_on_success(
+        self,
+        call_racf_mock: Mock,
+    ):
+        user_admin = UserAdmin(debug=True, additional_secret_traits=["omvs:uid"])
+        call_racf_mock.side_effect = [
+            TestUserConstants.TEST_EXTRACT_USER_RESULT_BASE_ONLY_SUCCESS_XML,
+            TestUserConstants.TEST_ALTER_USER_RESULT_EXTENDED_SUCCESS_XML,
+        ]
+        stdout = io.StringIO()
+        with contextlib.redirect_stdout(stdout):
+            user_admin.alter(
+                "squidwrd",
+                traits=TestUserConstants.TEST_ALTER_USER_REQUEST_TRAITS_EXTENDED,
+            )
+        success_log = self.ansi_escape.sub("", stdout.getvalue())
+        self.assertEqual(
+            success_log,
+            TestCommonConstants.TEST_ALTER_USER_ADDITIONAL_SECRET_ADDED_SUCCESS_LOG,
+        )
+        self.assertNotIn(
+            str(TestUserConstants.TEST_ALTER_USER_REQUEST_TRAITS_EXTENDED["omvs:uid"]),
+            success_log,
+        )
+
+    @patch("pyracf.common.irrsmo00.IRRSMO00.call_racf")
+    def test_alter_user_request_debug_log_additional_secret_added_get_redacted_on_error(
+        self,
+        call_racf_mock: Mock,
+    ):
+        user_admin = UserAdmin(debug=True, additional_secret_traits=["omvs:uid"])
+        call_racf_mock.side_effect = [
+            TestUserConstants.TEST_EXTRACT_USER_RESULT_BASE_ONLY_SUCCESS_XML,
+            TestUserConstants.TEST_ALTER_USER_RESULT_ERROR_XML,
+        ]
+        stdout = io.StringIO()
+        with contextlib.redirect_stdout(stdout):
+            try:
+                user_admin.alter(
+                    "squidwrd",
+                    traits=TestUserConstants.TEST_ALTER_USER_REQUEST_TRAITS_UID_ERROR,
+                )
+            except SecurityRequestError:
+                pass
+        error_log = self.ansi_escape.sub("", stdout.getvalue())
+        self.assertEqual(
+            error_log,
+            TestCommonConstants.TEST_ALTER_USER_ADDITIONAL_SECRET_ADDED_ERROR_LOG,
+        )
+        self.assertNotIn(
+            str(TestUserConstants.TEST_ALTER_USER_REQUEST_TRAITS_UID_ERROR["omvs:uid"]),
+            error_log,
+        )
+
     # ============================================================================
     # Group Administration
     # ============================================================================
@@ -180,6 +236,61 @@ class TestAdditionalSecretsResultRedaction(unittest.TestCase):
         self.assertNotIn(
             str(TestGroupConstants.TEST_ALTER_GROUP_REQUEST_TRAITS["omvs:gid"]),
             str(result),
+        )
+
+    @patch("pyracf.common.irrsmo00.IRRSMO00.call_racf")
+    def test_alter_group_request_debug_log_additional_secret_added_get_redacted_on_success(
+        self,
+        call_racf_mock: Mock,
+    ):
+        group_admin = GroupAdmin(debug=True, additional_secret_traits=["omvs:gid"])
+        call_racf_mock.side_effect = [
+            TestGroupConstants.TEST_EXTRACT_GROUP_RESULT_BASE_ONLY_SUCCESS_XML,
+            TestGroupConstants.TEST_ALTER_GROUP_RESULT_SUCCESS_XML,
+        ]
+        stdout = io.StringIO()
+        with contextlib.redirect_stdout(stdout):
+            group_admin.alter(
+                "testgrp0",
+                traits=TestGroupConstants.TEST_ALTER_GROUP_REQUEST_TRAITS,
+            )
+        success_log = self.ansi_escape.sub("", stdout.getvalue())
+        self.assertEqual(
+            success_log,
+            TestCommonConstants.TEST_ALTER_GROUP_ADDITIONAL_SECRET_ADDED_SUCCESS_LOG,
+        )
+        self.assertNotIn(
+            str(TestGroupConstants.TEST_ALTER_GROUP_REQUEST_TRAITS["omvs:gid"]),
+            success_log,
+        )
+
+    @patch("pyracf.common.irrsmo00.IRRSMO00.call_racf")
+    def test_alter_group_request_debug_log_additional_secret_added_get_redacted_on_error(
+        self,
+        call_racf_mock: Mock,
+    ):
+        group_admin = GroupAdmin(debug=True, additional_secret_traits=["omvs:gid"])
+        call_racf_mock.side_effect = [
+            TestGroupConstants.TEST_EXTRACT_GROUP_RESULT_BASE_ONLY_SUCCESS_XML,
+            TestGroupConstants.TEST_ALTER_GROUP_RESULT_ERROR_XML,
+        ]
+        stdout = io.StringIO()
+        with contextlib.redirect_stdout(stdout):
+            try:
+                group_admin.alter(
+                    "testgrp0",
+                    traits=TestGroupConstants.TEST_ALTER_GROUP_REQUEST_ERROR_TRAITS,
+                )
+            except SecurityRequestError:
+                pass
+        error_log = self.ansi_escape.sub("", stdout.getvalue())
+        self.assertEqual(
+            error_log,
+            TestCommonConstants.TEST_ALTER_GROUP_ADDITIONAL_SECRET_ADDED_ERROR_LOG,
+        )
+        self.assertNotIn(
+            str(TestGroupConstants.TEST_ALTER_GROUP_REQUEST_ERROR_TRAITS["omvs:gid"]),
+            error_log,
         )
 
     # ============================================================================
@@ -277,6 +388,62 @@ class TestAdditionalSecretsResultRedaction(unittest.TestCase):
             str(result),
         )
 
+    @patch("pyracf.common.irrsmo00.IRRSMO00.call_racf")
+    def test_alter_resource_request_debug_log_additional_secret_added_get_redacted_on_success(
+        self,
+        call_racf_mock: Mock,
+    ):
+        resource_admin = ResourceAdmin(
+            debug=True, additional_secret_traits=["base:owner"]
+        )
+        call_racf_mock.side_effect = [
+            TestResourceConstants.TEST_EXTRACT_RESOURCE_RESULT_BASE_SUCCESS_XML,
+            TestResourceConstants.TEST_ALTER_RESOURCE_RESULT_SUCCESS_XML,
+        ]
+        stdout = io.StringIO()
+        with contextlib.redirect_stdout(stdout):
+            resource_admin.alter(
+                "TESTING",
+                "ELIJTEST",
+                traits=TestResourceConstants.TEST_ALTER_RESOURCE_REQUEST_TRAITS,
+            )
+        success_log = self.ansi_escape.sub("", stdout.getvalue())
+        self.assertEqual(
+            success_log,
+            TestCommonConstants.TEST_ALTER_RESOURCE_ADDITIONAL_SECRET_ADDED_SUCCESS_LOG,
+        )
+        # No AssertNotInTest because Equal test to log is sufficient and value technically appears
+
+    @patch("pyracf.common.irrsmo00.IRRSMO00.call_racf")
+    def test_alter_resource_request_debug_log_additional_secret_added_get_redacted_on_error(
+        self,
+        call_racf_mock: Mock,
+    ):
+        secret_trait = "base:universal_access"
+        resource_admin = ResourceAdmin(
+            debug=True, additional_secret_traits=[secret_trait]
+        )
+        call_racf_mock.side_effect = [
+            TestResourceConstants.TEST_EXTRACT_RESOURCE_RESULT_BASE_SUCCESS_XML,
+            TestResourceConstants.TEST_ALTER_RESOURCE_RESULT_ERROR_XML,
+        ]
+        stdout = io.StringIO()
+        with contextlib.redirect_stdout(stdout):
+            try:
+                resource_admin.alter(
+                    "TESTING",
+                    "ELIJTEST",
+                    traits=TestResourceConstants.TEST_ALTER_RESOURCE_REQUEST_ERROR_TRAITS,
+                )
+            except SecurityRequestError:
+                pass
+        error_log = self.ansi_escape.sub("", stdout.getvalue())
+        self.assertEqual(
+            error_log,
+            TestCommonConstants.TEST_ALTER_RESOURCE_ADDITIONAL_SECRET_ADDED_ERROR_LOG,
+        )
+        # No AssertNotInTest because Equal test to log is sufficient and value technically appears
+
     # ============================================================================
     # Data Set Profile Administration
     # ============================================================================
@@ -341,6 +508,63 @@ class TestAdditionalSecretsResultRedaction(unittest.TestCase):
             str(result),
         )
 
+    @patch("pyracf.common.irrsmo00.IRRSMO00.call_racf")
+    def test_alter_data_set_request_debug_log_additional_secret_added_get_redacted_on_success(
+        self,
+        call_racf_mock: Mock,
+    ):
+        data_set_admin = DataSetAdmin(
+            debug=True, additional_secret_traits=["base:owner"]
+        )
+        call_racf_mock.side_effect = [
+            TestDataSetConstants.TEST_EXTRACT_DATA_SET_RESULT_BASE_ONLY_SUCCESS_XML,
+            TestDataSetConstants.TEST_ALTER_DATA_SET_RESULT_SUCCESS_XML,
+        ]
+        stdout = io.StringIO()
+        with contextlib.redirect_stdout(stdout):
+            data_set_admin.alter(
+                "ESWIFT.TEST.T1136242.P3020470",
+                traits=TestDataSetConstants.TEST_ALTER_DATA_SET_REQUEST_TRAITS,
+            )
+        success_log = self.ansi_escape.sub("", stdout.getvalue())
+        self.assertEqual(
+            success_log,
+            TestCommonConstants.TEST_ALTER_DATA_SET_ADDITIONAL_SECRET_ADDED_SUCCESS_LOG,
+        )
+        # No AssertNotInTest because Equal test to log is sufficient and value technically appears
+
+    @patch("pyracf.common.irrsmo00.IRRSMO00.call_racf")
+    def test_alter_data_set_request_debug_log_additional_secret_added_get_redacted_on_error(
+        self,
+        call_racf_mock: Mock,
+    ):
+        secret_trait = "base:universal_access"
+        data_set_admin = DataSetAdmin(
+            debug=True, additional_secret_traits=[secret_trait]
+        )
+        call_racf_mock.side_effect = [
+            TestDataSetConstants.TEST_EXTRACT_DATA_SET_RESULT_BASE_ONLY_SUCCESS_XML,
+            TestDataSetConstants.TEST_ALTER_DATA_SET_RESULT_ERROR_XML,
+        ]
+        stdout = io.StringIO()
+        with contextlib.redirect_stdout(stdout):
+            try:
+                data_set_admin.alter(
+                    "ESWIFT.TEST.T1136242.P3020470",
+                    traits=TestDataSetConstants.TEST_ALTER_DATA_SET_REQUEST_ERROR_TRAITS,
+                )
+            except SecurityRequestError:
+                pass
+        error_log = self.ansi_escape.sub("", stdout.getvalue())
+        self.assertEqual(
+            error_log,
+            TestCommonConstants.TEST_ALTER_DATA_SET_ADDITIONAL_SECRET_ADDED_ERROR_LOG,
+        )
+        # No AssertNotInTest because Equal test to log is sufficient and value technically appears
+
+    # ============================================================================
+    # Generic Testing for all Admin Types
+    # ============================================================================
     def test_incompatible_admin_custom_secret_redaction_error(self):
         admin_types = [
             (AccessAdmin, "permission"),
@@ -382,68 +606,3 @@ class TestAdditionalSecretsResultRedaction(unittest.TestCase):
                 f"Cannot add specified additional secrets to '{profile_type}' object.\n"
                 + incompatible_traits_text,
             )
-
-
-@patch("pyracf.common.irrsmo00.IRRSMO00.call_racf")
-class TestAdditionalSecretsLoggingRedaction(unittest.TestCase):
-    maxDiff = None
-    user_admin = UserAdmin(debug=True)
-    ansi_escape = re.compile(r"\x1B(?:[@-Z\\-_]|\[[0-?]*[ -/]*[@-~])")
-
-    # ============================================================================
-    # Add Additional Secrets
-    # ============================================================================
-    def test_alter_user_request_debug_log_additional_secret_added_get_redacted_on_success(
-        self,
-        call_racf_mock: Mock,
-    ):
-        user_admin = UserAdmin(debug=True, additional_secret_traits=["omvs:uid"])
-        call_racf_mock.side_effect = [
-            TestUserConstants.TEST_EXTRACT_USER_RESULT_BASE_ONLY_SUCCESS_XML,
-            TestUserConstants.TEST_ALTER_USER_RESULT_EXTENDED_SUCCESS_XML,
-        ]
-        stdout = io.StringIO()
-        with contextlib.redirect_stdout(stdout):
-            user_admin.alter(
-                "squidwrd",
-                traits=TestUserConstants.TEST_ALTER_USER_REQUEST_TRAITS_EXTENDED,
-            )
-        success_log = self.ansi_escape.sub("", stdout.getvalue())
-        self.assertEqual(
-            success_log,
-            TestCommonConstants.TEST_ALTER_USER_ADDITIONAL_SECRET_ADDED_SUCCESS_LOG,
-        )
-        self.assertNotIn(
-            str(TestUserConstants.TEST_ALTER_USER_REQUEST_TRAITS_EXTENDED["omvs:uid"]),
-            success_log,
-        )
-
-    # Secret redacted from command image but not from resulting error message.
-    # Marked experimental until resolved
-    def test_alter_user_request_debug_log_additional_secret_added_get_redacted_on_error(
-        self,
-        call_racf_mock: Mock,
-    ):
-        user_admin = UserAdmin(debug=True, additional_secret_traits=["omvs:uid"])
-        call_racf_mock.side_effect = [
-            TestUserConstants.TEST_EXTRACT_USER_RESULT_BASE_ONLY_SUCCESS_XML,
-            TestUserConstants.TEST_ALTER_USER_RESULT_ERROR_XML,
-        ]
-        stdout = io.StringIO()
-        with contextlib.redirect_stdout(stdout):
-            try:
-                user_admin.alter(
-                    "squidwrd",
-                    traits=TestUserConstants.TEST_ALTER_USER_REQUEST_TRAITS_UID_ERROR,
-                )
-            except SecurityRequestError:
-                pass
-        error_log = self.ansi_escape.sub("", stdout.getvalue())
-        self.assertEqual(
-            error_log,
-            TestCommonConstants.TEST_ALTER_USER_ADDITIONAL_SECRET_ADDED_ERROR_LOG,
-        )
-        self.assertNotIn(
-            str(TestUserConstants.TEST_ALTER_USER_REQUEST_TRAITS_UID_ERROR["omvs:uid"]),
-            error_log,
-        )
--- a/tests/group/test_group_debug_logging.py
+++ b/tests/group/test_group_debug_logging.py
@@ -35,7 +35,7 @@ class TestGroupDebugLogging(unittest.TestCase):
         stdout = io.StringIO()
         with contextlib.redirect_stdout(stdout):
             self.group_admin.alter(
-                "TESTGRP0", traits=TestGroupConstants.TEST_ALTER_GROUP_REQUEST_TRAITS
+                "testgrp0", traits=TestGroupConstants.TEST_ALTER_GROUP_REQUEST_TRAITS
             )
         success_log = self.ansi_escape.sub("", stdout.getvalue())
         self.assertEqual(success_log, TestGroupConstants.TEST_ALTER_GROUP_SUCCESS_LOG)
@@ -52,7 +52,7 @@ class TestGroupDebugLogging(unittest.TestCase):
         with contextlib.redirect_stdout(stdout):
             try:
                 self.group_admin.alter(
-                    "TESTGRP0",
+                    "testgrp0",
                     traits=TestGroupConstants.TEST_ALTER_GROUP_REQUEST_ERROR_TRAITS,
                 )
             except SecurityRequestError:
@@ -72,7 +72,7 @@ class TestGroupDebugLogging(unittest.TestCase):
         )
         stdout = io.StringIO()
         with contextlib.redirect_stdout(stdout):
-            self.group_admin.extract("TESTGRP0", segments=["omvs"])
+            self.group_admin.extract("testgrp0", segments=["omvs"])
         success_log = self.ansi_escape.sub("", stdout.getvalue())
         self.assertEqual(
             success_log, TestGroupConstants.TEST_EXTRACT_GROUP_BASE_OMVS_SUCCESS_LOG
@@ -88,7 +88,7 @@ class TestGroupDebugLogging(unittest.TestCase):
         stdout = io.StringIO()
         with contextlib.redirect_stdout(stdout):
             try:
-                self.group_admin.extract("TESTGRP0", segments=["omvs"])
+                self.group_admin.extract("testgrp0", segments=["omvs"])
             except SecurityRequestError:
                 pass
         error_log = self.ansi_escape.sub("", stdout.getvalue())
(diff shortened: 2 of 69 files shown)


pyracf/common/utilities/logger.py
```diff
@@ -149,19 +149,33 @@ def __redact_request_dictionary(
     def __redact_string(
         self,
         input_string: Union[str, bytes],
-        start_ind: int,
-        end_pattern: Union[str, bytes],
-    ):
+        key: str,
```
lcarcaramo, line 152: Change this to `trait_key` maybe for better clarity?

pyracf/common/utilities/logger.py
```diff
@@ -149,19 +149,33 @@ def __redact_request_dictionary(
...
+                rf"{key.upper()}( +)\(\'.*?(?<!\')\'\)",
+                rf"{key.upper()}\1('{asterisks}')",
+                input_string,
+            )
+        else:
+            input_string = re.sub(
+                rf"{key.upper()}( +)\(.*?(?<!\\)\)",
+                rf"{key.upper()}\1({asterisks})",
+                input_string,
+            )
```
lcarcaramo, line 175: Are you able to provide an explanation of how these regex expressions work?

pyracf/common/utilities/logger.py
```diff
@@ -203,27 +221,38 @@ def redact_result_xml(
...
+                    + r"REVIEW DOCUMENTATION OF \1 FOR MORE INFORMATION</message>",
+                    security_result.decode("cp1047"),
+                ).encode("cp1047")
+            else:
+                security_result = re.sub(
+                    rf"<message>([A-Z]*[0-9]*[A-Z]) [^<>]*{racf_key.upper()}[^<>]*<\/message>",
+                    rf"<message>REDACTED MESSAGE CONCERNING {racf_key.upper()}, "
+                    + r"REVIEW DOCUMENTATION OF \1 FOR MORE INFORMATION</message>",
+                    security_result,
+                )
```
lcarcaramo, line 255: Also explain how these regexes work.

tests/user/test_user_result_parser.py
```diff
@@ -523,3 +523,27 @@ def test_user_admin_custom_secret_redacted_on_error(
...
+        self.assertEqual(
+            result,
+            TestUserConstants.TEST_ALTER_USER_RESULT_SUCCESS_INST_DATA_SECRET_DICTIONARY,
+        )
+        self.assertNotIn(
+            TestUserConstants.TEST_ALTER_USER_REQUEST_TRAITS_INST_DATA[
+                "base:installation_data"
+            ],
+            result,
+        )
```
lcarcaramo, line 549: I wonder if there should be an extract test too for the complex installation data?

pyracf/common/utilities/logger.py
ElijahSwiftIBM: sure! Done

tests/user/test_user_result_parser.py
ElijahSwiftIBM: I actually am working on stabilizing this more fully in my RAS branch. There is a new test but it required new extract logic to actually work.

ElijahSwiftIBM: I would still not consider INSTALLATION_DATA "Stable" at the end of this branch, but once this is approved, I would consider additional_secrets_redaction "stable".

pyracf/common/utilities/logger.py
```diff
@@ -149,30 +149,97 @@ def __redact_request_dictionary(
...
+        is_bytes = False
         if isinstance(input_string, bytes):
-            asterisks = asterisks.encode("cp1047")
-        pre_keyword = input_string[:start_ind]
-        post_keyword = end_pattern.join(input_string[start_ind:].split(end_pattern)[1:])
-        return pre_keyword + asterisks + end_pattern + post_keyword
+            input_string = input_string.decode("cp1047")
+            is_bytes = True
+        quoted = re.search(
+            rf"{trait_key.upper()}( +){{0,}}\(\'.*?(?<!\')\'\)", input_string
```
lcarcaramo, line 196: Some of the regexes are repeated multiple times. It might make sense to initialize these regexes as variables at the beginning. This will not only decrease repetition, which reduces the likelihood that any instance of a repeated value has a typo that is more likely to get overlooked due to duplication, but it will also increase the signal to noise ratio of the code that follows making it easier to read and understand.

```python
asterisks = "********"
is_bytes = False
quoted_regex = rf"{trait_key.upper()}( +){{0,}}\(\'.*?(?<!\')\'\)"
nested_regex = rf"{trait_key.upper()}( +){{0,}}\([^)]*\(.*\)( +){{0,}}\)"
default_regex = rf"{trait_key.upper()}( +){{0,}}\(.*?(?<!\\)\)"
redacted_regex = rf"{trait_key.upper()}\1({asterisks})"
```

pyracf/common/utilities/logger.py
```diff
@@ -149,30 +149,97 @@ def __redact_request_dictionary(
...
+        "end pattern" of ) and ) separated by a variable (potentially zero) whitespace.
+
+        If neither of these two patterns is found for a supplied trait_key, then it is assumed
+        this trait is set with the default pattern below.
+
+        Regex 3 ("default") - {trait_key.upper()}( +){{0,}}\(.*?(?<!\\)\)
+        This is designed to match the pattern TRAIT(value) by matching the TRAIT name, a variable
+        (potentially zero) amount of spaces, then the (value) portion which must start and end
+        with ( and ), but can conceivably contain any characters, but a negative lookbehind
+        is used to look for any escape \ character that would indicate the matching of the
```
lcarcaramo, line 183: Is `\` a typo? Is this sentence saying that the regex can handle extraneous parenthesis between the `(` and `)`?

pyracf/common/utilities/logger.py
```diff
@@ -149,30 +149,97 @@ def __redact_request_dictionary(
...
+        variable (potentially zero) amount of spaces, then the ('value') portion which must
+        start and end with (' and '), but can conceivably contain any characters, but a negative
+        lookbehind is used to look for any unescaped single quotes that would indicate the matching
+        of ') is otherwise a coincidence.
+
+        Regex 2 ("nested") - {trait_key.upper()}( +){{0,}}\([^)]*\(.*\)( +){{0,}}\)
+        This is designed to match the pattern TRAIT( subtrait1(value) subtrait2(value)) by
+        matching the TRAIT name, a variable (potentially zero) amount of spaces, then the
+        ( subtrait1(value) subtrait2(value)) portion which must start and end with ( and ),
+        but must also contain a nested set of opened parentheses rather than a direct seqence of
```
lcarcaramo, line 171: Do you mean to say `nested set of open and close parentheses` here? `open parentheses` implies two open parentheses `((`.

pyracf/common/utilities/logger.py
```diff
@@ -149,30 +149,97 @@ def __redact_request_dictionary(
...
+            return input_string.encode("cp1047")
+        return input_string
 
     def redact_request_xml(
         self,
         xml_string: Union[str, bytes],
         secret_traits: dict,
     ) -> Union[str, bytes]:
-        """
+        r"""
```
lcarcaramo, line 229: Is the `r` necessary to redact the regex stuff here?

pyracf/common/utilities/logger.py
```diff
@@ -149,30 +149,97 @@ def __redact_request_dictionary(
...
         Based the following xml pattern:
             '<xmltag attribute="any">xml value</xmltag>'
         This function also accounts for any number of arbitrary xml attributes.
+
+        This function employs the following regular expression:
+        {xml_key}(.*)>.*<\/{xml_key} - Designed to match the above pattern by starting and ending
+        with the xmltag string as shown, but the starting tage allows for any characters between
+        "xmltag" and the > character to allow for the attribute specification shown above. This
+        results in the starting of the xml as {xml_key}(.*)> and the ending as <\/{xml_key}.
+        The miscellaneous characters are "captured" as a variable and preserved by the
```
lcarcaramo, line 240: Change to: `The characters between the xml tags are "captured"`.

pyracf/common/utilities/logger.py
```diff
@@ -149,30 +149,97 @@ def __redact_request_dictionary(
...
-        """
+        r"""
         Redact a list of specific secret traits in a request xml string or bytes object.
         Based the following xml pattern:
             '<xmltag attribute="any">xml value</xmltag>'
         This function also accounts for any number of arbitrary xml attributes.
+
+        This function employs the following regular expression:
+        {xml_key}(.*)>.*<\/{xml_key} - Designed to match the above pattern by starting and ending
+        with the xmltag string as shown, but the starting tage allows for any characters between
```
lcarcaramo, line 237: Typo `tage`.

pyracf/common/utilities/logger.py
```diff
@@ -191,7 +258,11 @@ def redact_request_xml(
             # <tag operation="del" />
             if f"</{xml_key}>" not in xml_string:
                 continue
-            xml_string = self.__redact_string(xml_string, match.end(), f"</{xml_key}")
+            xml_string = re.sub(
```
lcarcaramo, line 261: Is this still necessary?

```python
match = re.search(rf"\<{xml_key}+[^>]*\>", xml_string)
if not match:
    continue
```

lcarcaramo: A brief explanation of it's purposes might be helpful.

pyracf/common/utilities/logger.py
```diff
@@ -191,7 +258,11 @@ def redact_request_xml(
...
+                rf"{xml_key}(.*)>.*<\/{xml_key}",
+                rf"{xml_key}\1>********</{xml_key}",
+                xml_string,
+            )
```
lcarcaramo, line 265: Initialize regexes at the beginning for clarity and to increase signal to noise ratio?

```python
is_bytes = False
? = rf"\<{xml_key}+[^>]*\>"
start_tag_end_tag_regex = rf"{xml_key}(.*)>.*<\/{xml_key}"
redacted_regex = rf"{xml_key}\1>********</{xml_key}"
```

pyracf/common/utilities/logger.py
```diff
@@ -200,30 +271,49 @@ def redact_result_xml(
...
+            "TRAIT ('value')"
         This function also accounts for varied amounts of whitespace in the pattern.
         """
         if isinstance(security_result, list):
             return security_result
         for xml_key in secret_traits.values():
             racf_key = xml_key.split(":")[1] if ":" in xml_key else xml_key
-            end_pattern = ")"
+            if racf_key in racf_key_map:
+                print(racf_key_map[racf_key])
```
lcarcaramo, line 290: Extraneous `print`.

pyracf/common/utilities/logger.py
```diff
@@ -200,30 +271,49 @@ def redact_result_xml(
...
+                    rf"<message>([A-Z]*[0-9]*[A-Z]) [^<>]*{racf_key.upper()}[^<>]*<\/message>",
+                    rf"<message>REDACTED MESSAGE CONCERNING {racf_key.upper()}, "
+                    + r"REVIEW DOCUMENTATION OF \1 FOR MORE INFORMATION</message>",
+                    security_result.decode("cp1047"),
+                ).encode("cp1047")
+            else:
+                security_result = re.sub(
+                    rf"<message>([A-Z]*[0-9]*[A-Z]) [^<>]*{racf_key.upper()}[^<>]*<\/message>",
+                    rf"<message>REDACTED MESSAGE CONCERNING {racf_key.upper()}, "
+                    + r"REVIEW DOCUMENTATION OF \1 FOR MORE INFORMATION</message>",
```
lcarcaramo, line 314: Again, may make sense to initialize repeated regex at the beginning of the function:

```python
racf_command_argument_regex = rf"{racf_key.upper()}( +){{0,}}\("
racf_message_regex = rf"<message>([A-Z]*[0-9]*[A-Z]) [^<>]*{racf_key.upper()}[^<>]*<\/message>"
racf_redacted_message_regex = (
    rf"<message>REDACTED MESSAGE CONCERNING {racf_key.upper()}, "
    + r"REVIEW DOCUMENTATION OF \1 FOR MORE INFORMATION</message>"
)
```

pyracf/common/security_admin.py
```diff
@@ -54,6 +54,10 @@ class SecurityAdmin:
 
     _valid_segment_traits = {}
     _extracted_key_value_pair_segment_traits_map = {}
+    _racf_key_map = {}
+    _racf_message_key_map = {}
+    _racf_key_map = {}
+    _racf_message_key_map = {}
```
lcarcaramo, line 60: These definitions are repeated. My guess this is a typo?

lcarcaramo, line 60: Combine `_racf_key_map` and `_racf_message_key_map` into single structure? This reduces the number of arguments that need to get passed around by callers/users of our redaction function.

```python
_racf_trait_and_message_key_map = {}
```

```python
self._racf_trait_and_message_map = {
    "trait_map": {
        "audaltr": "audit",
        "audcntl": "audit",
        "audnone": "audit",
        "audread": "audit",
        "audupdt": "audit",
    },
    "message_map": {"uacc": "universal access"}
}
```

pyracf/common/security_admin.py
```diff
@@ -54,6 +54,10 @@ class SecurityAdmin:
 
     _valid_segment_traits = {}
     _extracted_key_value_pair_segment_traits_map = {}
+    _racf_key_map = {}
```
lcarcaramo, line 57: Add brief comment indicating that these structures are used for redaction?

tests/common/test_additional_secrets_redaction.py
```diff
@@ -0,0 +1,345 @@
...
+            exception.exception.result,
+            TestCommonConstants.TEST_ALTER_USER_RESULT_ERROR_UID_SECRET_DICTIONARY,
+        )
+        self.assertNotIn(
+            TestUserConstants.TEST_ALTER_USER_REQUEST_TRAITS_UID_ERROR["omvs:uid"],
+            exception.exception.result,
+        )
+
+    @patch("pyracf.common.irrsmo00.IRRSMO00.call_racf")
+    def test_user_admin_custom_secret_redacted_when_complex_characters(
```
lcarcaramo, line 79: Is there no failure scenario to test for installation data?

tests/common/test_additional_secrets_redaction.py
```diff
@@ -0,0 +1,345 @@
...
+            result,
+            TestCommonConstants.TEST_ALTER_RESOURCE_RESULT_SUCCESS_OWNER_SECRET_DICTIONARY,
+        )
+        self.assertNotIn(
+            TestResourceConstants.TEST_ALTER_RESOURCE_REQUEST_TRAITS["base:owner"],
+            result,
+        )
+
+    @patch("pyracf.common.irrsmo00.IRRSMO00.call_racf")
+    def test_resource_admin_custom_mapped_secret_redacted_on_success(
```
lcarcaramo, line 212: Does it make sense to test an error scenario for audit rules?

tests/common/test_additional_secrets_redaction.py
```diff
@@ -0,0 +1,345 @@
...
+        )
+
+    def test_user_admin_custom_secret_redacted_request(self):
+        user_admin = UserAdmin(
+            additional_secret_traits=["omvs:uid"], generate_requests_only=True
+        )
+        result = user_admin.alter(
+            "squidwrd",
+            traits=TestUserConstants.TEST_ALTER_USER_REQUEST_TRAITS_EXTENDED,
+        )
```
lcarcaramo, line 110: Should we also be comparing to request xml samples? I believe we do this elsewhere.

tests/user/user_log_samples/alter_user_additional_secret_added_error.log
```diff
@@ -271,9 +271,9 @@
...
           "messages": [
-            "IKJ56702I INVALID UID, 90000000000",
-            "IKJ56701I MISSING OMVS UID+",
-            "IKJ56701I MISSING OMVS USER ID (UID), 1-10 NUMERIC DIGITS"
+            "REDACTED MESSAGE CONCERNING UID, REVIEW DOCUMENTATION OF IKJ56702I FOR MORE INFORMATION",
+            "REDACTED MESSAGE CONCERNING UID, REVIEW DOCUMENTATION OF IKJ56701I FOR MORE INFORMATION",
+            "REDACTED MESSAGE CONCERNING UID, REVIEW DOCUMENTATION OF IKJ56701I FOR MORE INFORMATION"
           ]
         }
       ]
```
lcarcaramo, line 279: Does this still belong in user admin anymore? Should all secret redaction be tested in common since it is a common feature?

tests/user/test_user_constants.py
```diff
@@ -218,6 +212,10 @@ def get_sample(sample_file: str) -> Union[str, bytes]:
 )
 TEST_ALTER_USER_REQUEST_TRAITS_UID_ERROR = dict(TEST_ALTER_USER_REQUEST_TRAITS_EXTENDED)
 TEST_ALTER_USER_REQUEST_TRAITS_UID_ERROR["omvs:uid"] = 90000000000
+TEST_ALTER_USER_REQUEST_TRAITS_INST_DATA = dict(TEST_ALTER_USER_REQUEST_TRAITS_EXTENDED)
+TEST_ALTER_USER_REQUEST_TRAITS_INST_DATA["base:installation_data"] = (
+    "Test = Value; Other(stuff goes here)'')"
+)
```
lcarcaramo, line 218: Does this belong in common?

tests/common/test_additional_secrets_redaction.py
```diff
@@ -0,0 +1,345 @@
...
+        self.assertEqual(
+            result,
+            TestCommonConstants.TEST_ALTER_USER_RESULT_SUCCESS_UID_SECRET_DICTIONARY,
+        )
+        self.assertNotIn(
+            TestUserConstants.TEST_ALTER_USER_REQUEST_TRAITS_EXTENDED["omvs:uid"],
+            result,
```
lcarcaramo, line 51: Not sure if these `assertNotIn` assertions actually work correctly since result is a dictionary. You might need to stringify the result dictionary in order for this to work.

pyracf/common/security_admin.py
```diff
@@ -170,13 +174,21 @@ def __raw_dump(self) -> None:
...
+            )
 
     # ============================================================================
     # Secrets Redaction
     # ============================================================================
     def __add_additional_secret_traits(self, additional_secret_traits: list) -> None:
         """Add additional fields to be redacted in logger output."""
+        unsupported_profile_types = ["permission", "groupConnection", "systemSettings"]
+        if self._profile_type in unsupported_profile_types:
+            return
```
lcarcaramo, line 191: I wonder if we should be raising an exception if the additional secret the user provides is bad?

pyracf/common/utilities/logger.py
ElijahSwiftIBM: Excellent points, I had to add a regex for `redacted_w_quotes`, but otherwise adopted this.

pyracf/common/utilities/logger.py
ElijahSwiftIBM: I was trying to indicate that \ was the escape character targeted by this negative lookbehind, but I was worried that extraneous characters like referring to it as '\' or `\` would suggest other characters are involved. I will work to re-word this explanation.

pyracf/common/utilities/logger.py
ElijahSwiftIBM: I think that is clearer. I have changed it.

pyracf/common/utilities/logger.py
ElijahSwiftIBM: without the "r", some of the characters used in this string like `\` take on new meaning and break the string. Treating it as a raw string prevents that sort of interpretation of special characters.

pyracf/common/utilities/logger.py
ElijahSwiftIBM: Change made.

pyracf/common/utilities/logger.py
ElijahSwiftIBM: good catch.

pyracf/common/utilities/logger.py
ElijahSwiftIBM: no, this is no longer necessary, my mistake. It has been removed.

pyracf/common/utilities/logger.py
ElijahSwiftIBM: Yes, the initial tag regex is also no longer necessary. Since these use the xml_key from this for loop, the definitions could not go all the way back to where is_bytes is defined.

pyracf/common/utilities/logger.py
ElijahSwiftIBM: Sorry, I thought I had gotten rid of those.

pyracf/common/utilities/logger.py
ElijahSwiftIBM: Yup, all true, but still good to build these once and re-use them.

pyracf/common/security_admin.py
ElijahSwiftIBM: yes, my apologies. Issue when debugging, no longer a problem.

ElijahSwiftIBM: Took a bit of refactoring, but that is all functional now. Good idea.

pyracf/common/security_admin.py
ElijahSwiftIBM: Sure.

tests/common/test_additional_secrets_redaction.py
ElijahSwiftIBM: I did not add one, but the redaction of secrets is functionally no different on success cases or error cases with the potential exception of messages concerning the field. I am not aware of any specific messages to INSTALLATION DATA. I can add more test cases when formally stabilizing Installation data.

tests/user/user_log_samples/alter_user_additional_secret_added_error.log
ElijahSwiftIBM: Yes, my mistake, I forgot we did testing in user debug logging. I have moved this to common and will add comparable debug_logging testing for the other function groups as well.

tests/user/test_user_constants.py
ElijahSwiftIBM: Yes, it should be there, good catch.

tests/common/test_additional_secrets_redaction.py
ElijahSwiftIBM: That is an excellent point and has now been resolved.

tests/common/test_additional_secrets_redaction.py
ElijahSwiftIBM: You are correct, I was testing this in debug logging for user which had not been moved over. It now has been and I am adding similar tests for the other function groups.

pyracf/common/security_admin.py
ElijahSwiftIBM: That's a fair point, I will add an exception for this.

ElijahSwiftIBM:
Dealt with many of the comments here, but there is still more to do.

- `SecretsRedactionError` needs unit testing now that it exists
- More debug logging unit tests for the other admin functions

I will deal with these next week, but the rest is ready for review.

pyracf/common/utilities/logger.py
```diff
@@ -149,49 +149,118 @@ def __redact_request_dictionary(
...
+        the "end pattern" of ) and ) separated by a variable (potentially zero) whitespace.
+
+        If neither of these two patterns is found for a supplied trait_key, then it is assumed
+        this trait is set with the default pattern below.
+
+        Regex 3 ("default") - {trait_key.upper()}( +){{0,}}\(.*?(?<!\\)\)
+        This is designed to match the pattern TRAIT(value) by matching the TRAIT name, a variable
+        (potentially zero) amount of spaces, then the (value) portion which must start and end
+        with ( and ), but can conceivably contain any characters, but a negative lookbehind
+        is used to look for any escape character `\` that would indicate the matching of the
```
lcarcaramo, line 183: Surround `\` with `'` characters instead. Only use the back tick character for markdown.

pyracf/common/utilities/logger.py
```diff
@@ -149,49 +149,118 @@ def __redact_request_dictionary(
...
         """
         is_bytes = False
         if isinstance(xml_string, bytes):
             is_bytes = True
             xml_string = xml_string.decode("utf-8")
         for xml_key in secret_traits.values():
-            match = re.search(rf"\<{xml_key}+[^>]*\>", xml_string)
-            if not match:
-                continue
+            start_tag_eng_tag_regex = rf"{xml_key}(.*)>.*<\/{xml_key}"
```
lcarcaramo, line 250: typo

tests/common/test_additional_secrets_redaction.py
```diff
@@ -0,0 +1,402 @@
...
+                traits=TestUserConstants.TEST_ALTER_USER_REQUEST_TRAITS_EXTENDED,
+            )
+        success_log = self.ansi_escape.sub("", stdout.getvalue())
+        self.assertEqual(
+            success_log,
+            TestCommonConstants.TEST_ALTER_USER_ADDITIONAL_SECRET_ADDED_SUCCESS_LOG,
+        )
+        self.assertNotIn(
+            TestUserConstants.TEST_ALTER_USER_REQUEST_TRAITS_EXTENDED["omvs:uid"],
+            success_log,
```
lcarcaramo, line 371: `TestUserConstants.TEST_ALTER_USER_REQUEST_TRAITS_EXTENDED["omvs:uid"]` should be casted to a string here, no?

tests/common/test_additional_secrets_redaction.py
```diff
@@ -0,0 +1,402 @@
...
+                )
+            except SecurityRequestError:
+                pass
+        error_log = self.ansi_escape.sub("", stdout.getvalue())
+        self.assertEqual(
+            error_log,
+            TestCommonConstants.TEST_ALTER_USER_ADDITIONAL_SECRET_ADDED_ERROR_LOG,
+        )
+        self.assertNotIn(
+            f"({TestUserConstants.TEST_ALTER_USER_REQUEST_TRAITS_UID_ERROR['omvs:uid']})",
```
lcarcaramo, line 400: `str(TestUserConstants.TEST_ALTER_USER_REQUEST_TRAITS_UID_ERROR['omvs:uid'])`?

tests/common/test_additional_secrets_redaction.py
```diff
@@ -0,0 +1,402 @@
...
+        result = user_admin.alter(
+            "squidwrd",
+            traits=TestUserConstants.TEST_ALTER_USER_REQUEST_TRAITS_EXTENDED,
+        )
+        self.assertEqual(
+            result,
+            TestCommonConstants.TEST_ALTER_USER_RESULT_SUCCESS_UID_SECRET_DICTIONARY,
+        )
+        self.assertNotIn(
+            TestUserConstants.TEST_ALTER_USER_REQUEST_TRAITS_EXTENDED["omvs:uid"],
```
lcarcaramo, line 53: Convert to string?

tests/common/test_additional_secrets_redaction.py
```diff
@@ -0,0 +1,402 @@
...
+        result = group_admin.alter(
+            "testgrp0",
+            traits=TestGroupConstants.TEST_ALTER_GROUP_REQUEST_TRAITS,
+        )
+        self.assertEqual(
+            result,
+            TestCommonConstants.TEST_ALTER_GROUP_RESULT_SUCCESS_GID_SECRET_DICTIONARY,
+        )
+        self.assertNotIn(
+            TestGroupConstants.TEST_ALTER_GROUP_REQUEST_TRAITS["omvs:gid"],
```
lcarcaramo, line 141: cast to string?

tests/common/test_additional_secrets_redaction.py
```diff
@@ -0,0 +1,402 @@
...
+            group_admin.alter(
+                "testgrp0",
+                traits=TestGroupConstants.TEST_ALTER_GROUP_REQUEST_ERROR_TRAITS,
+            )
+        self.assertEqual(
+            exception.exception.result,
+            TestCommonConstants.TEST_ALTER_GROUP_RESULT_ERROR_GID_SECRET_DICTIONARY,
+        )
+        self.assertNotIn(
+            TestGroupConstants.TEST_ALTER_GROUP_REQUEST_ERROR_TRAITS["omvs:gid"],
```
lcarcaramo, line 165: cast to string?

tests/common/test_additional_secrets_redaction.py
```diff
@@ -0,0 +1,402 @@
...
+    def test_group_admin_custom_secret_redacted_request(self):
+        group_admin = GroupAdmin(
+            additional_secret_traits=["omvs:gid"], generate_requests_only=True
+        )
+        result = group_admin.alter(
+            "squidwrd",
+            traits=TestGroupConstants.TEST_ALTER_GROUP_REQUEST_TRAITS,
+        )
+        self.assertNotIn(
+            TestGroupConstants.TEST_ALTER_GROUP_REQUEST_TRAITS["omvs:gid"],
```
lcarcaramo, line 178: cast to string?

tests/common/test_additional_secrets_redaction.py
```diff
@@ -0,0 +1,402 @@
...
+                "ELIJTEST",
+                traits=TestResourceConstants.TEST_ALTER_RESOURCE_REQUEST_ERROR_TRAITS,
+            )
+        self.assertEqual(
+            exception.exception.result,
+            TestCommonConstants.TEST_ALTER_RESOURCE_RESULT_ERROR_UACC_SECRET_DICTIONARY,
+        )
+        self.assertNotIn(
+            TestResourceConstants.TEST_ALTER_RESOURCE_REQUEST_ERROR_TRAITS[
+                f"{secret_trait}"
```
lcarcaramo, line 256: Change `f"{secret_trait}"` to `secret_trait`.

pyracf/common/security_admin.py
lcarcaramo: Does this need it's own exception, or do you think that we could just using `ValueError` here? I just want to avoid us having dozens of exceptions for every possible case where the user provides a bad value. We may also want to revisit some of the trait errors as well. May make sense to use `ValueError` in those scenarios.

pyracf/common/utilities/logger.py
ElijahSwiftIBM: Added the following:

        This function employs the following regular expression:
        <message>([A-Z]*[0-9]*[A-Z]) [^<>]*{racf_key.upper()}[^<>]*<\/message>" - 
        Designed to match the above pattern by starting and ending with the message xml tag 
        string as shown, the value of the message is targeted based on the racf_key. This
        should capture only messages that contain information about a redacted key.
        The message identifier is "captured" as a variable and preserved by the substitution
        operation through the use of the \1 supplied in the replacement string.

pyracf/common/utilities/logger.py
```diff
@@ -149,49 +149,118 @@ def __redact_request_dictionary(
...
+        start and end with (' and '), but can conceivably contain any characters, but a negative
+        lookbehind is used to look for any unescaped single quotes that would indicate the matching
+        of ') is otherwise a coincidence.
+
+        Regex 2 ("nested") - {trait_key.upper()}( +){{0,}}\([^)]*\(.*\)( +){{0,}}\)
+        This is designed to match the pattern TRAIT( subtrait1(value) subtrait2(value)) by
+        matching the TRAIT name, a variable (potentially zero) amount of spaces, then the
+        ( subtrait1(value) subtrait2(value)) portion which must start and end with ( and ),
+        but must also contain a nested set of open and close parentheses rather than a direct
+        seqence of them. The pattern looks for these nested open parenthesis as a sequence would
```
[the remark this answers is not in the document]
ElijahSwiftIBM: I changed to the following, please let me know if this is still unclear:

        Regex 2 ("nested") - {trait_key.upper()}( +){{0,}}\([^)]*\(.*\)( +){{0,}}\)
        This is designed to match the pattern TRAIT( subtrait1(value) subtrait2(value)) by
        matching the TRAIT name, a variable (potentially zero) amount of spaces, then the
        ( subtrait1(value) subtrait2(value)) portion which must start and end with ( and ),
        but must also contain a'(' before the ')'. This indicates that there is a "nested"
        structure rather than a sequential one. In the example, this portion of the pattern
        matches '( subtrait1(', but would not match '(value) subtrait2(' because of the ')'
        character between the two '(' characters. The pattern looks for these nested open
        parenthesis as a sequence would have a ')' character between them. Then the expression
        allows any non-newline characters to handle all possible trait values and subtrait names
        followed by the "end pattern" of ')' and ')' separated by a variable (potentially zero)
        whitespace.

pyracf/common/utilities/logger.py
ElijahSwiftIBM: noted and this syntax is now largely followed elsewhere in the comments.

tests/common/test_additional_secrets_redaction.py
ElijahSwiftIBM: good catch, resolved.

tests/common/test_additional_secrets_redaction.py
ElijahSwiftIBM: good catch, resolved.

tests/common/test_additional_secrets_redaction.py
ElijahSwiftIBM: good catch, resolved.

tests/common/test_additional_secrets_redaction.py
ElijahSwiftIBM: I'm not sure why I wrote this one this way.

tests/common/test_additional_secrets_redaction.py
ElijahSwiftIBM: good catch, resolved.

tests/common/test_additional_secrets_redaction.py
ElijahSwiftIBM: good catch, resolved.

tests/common/test_additional_secrets_redaction.py
```diff
@@ -0,0 +1,402 @@
...
+            data_set_admin.alter(
+                "ESWIFT.TEST.T1136242.P3020470",
+                traits=TestDataSetConstants.TEST_ALTER_DATA_SET_REQUEST_TRAITS,
+            )
+        self.assertEqual(
+            exception.exception.result,
+            TestCommonConstants.TEST_ALTER_DATA_SET_RESULT_ERROR_UACC_SECRET_DICTIONARY,
+        )
+        self.assertNotIn(
+            TestDataSetConstants.TEST_ALTER_DATA_SET_REQUEST_TRAITS[f"{secret_trait}"],
```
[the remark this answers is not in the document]
ElijahSwiftIBM: Again, not sure why I did that at all, but changed.

tests/common/test_additional_secrets_redaction.py
ElijahSwiftIBM: Again, not sure why I did that at all, but changed.

pyracf/common/security_admin.py
```diff
@@ -170,24 +172,49 @@ def __raw_dump(self) -> None:
...
+                f"\n'{self._profile_type}' object does not support additional secrets redaction."
+            )
+            raise ValueError(error_message)
+        bad_secret_traits = []
         for secret in additional_secret_traits:
             if secret in self.__secret_traits:
                 continue
             if ":" not in secret:
+                bad_secret_traits.append(
+                    f"\nCould not map {secret} to a valid segment trait."
```
lcarcaramo, line 201: I think it is sufficient to say `"\n'{secret}' is not a valid trait."`

pyracf/common/security_admin.py
```diff
@@ -170,24 +172,49 @@ def __raw_dump(self) -> None:
...
+            )
 
     # ============================================================================
     # Secrets Redaction
     # ============================================================================
     def __add_additional_secret_traits(self, additional_secret_traits: list) -> None:
         """Add additional fields to be redacted in logger output."""
+        unsupported_profile_types = ["permission", "groupConnection", "systemSettings"]
+        error_message = (
+            f"Cannot add specified additional secrets to '{self._profile_type}' object."
```
lcarcaramo, line 188: You should be able to use `inspect.stack()` here to get the name of the "admin" class (i.e., `UserAdmin`): https://github.com/ambitus/pyracf/blob/dd23a06a2a886d7d9b122be36dfdade155917c89/pyracf/common/utilities/logger.py#L116

Also change the word `object` to `type` and change `Cannot` to `Unable`.

You should get an exception message like this when you are done:

```console
Unable to add specified additional secrets to 'UserAdmin' type.
```

Make similar corresponding changes to the unsupported profile type exception message.

pyracf/common/utilities/logger.py
lcarcaramo: Just a slight working change suggestion:

```
Designed to match any message xml element that contains the provided racf_key.
```

Instead of
```
Designed to match the above pattern by starting and ending with the message xml tag string as shown, the value of the message is targeted based on the racf_key.
```

pyracf/common/utilities/logger.py
```diff
@@ -200,30 +273,62 @@ def redact_result_xml(
...
+            racf_command_argument_regex = rf"{racf_key.upper()}( +){{0,}}\("
             if isinstance(security_result, bytes):
                 match = re.search(
-                    rf"{racf_key.upper()} +\(", security_result.decode("cp1047")
+                    racf_command_argument_regex, security_result.decode("cp1047")
                 )
-                end_pattern = end_pattern.encode("cp1047")
             else:
-                match = re.search(rf"{racf_key.upper()} +\(", security_result)
+                match = re.search(rf"{racf_key.upper()}( +){{0,}}\(", security_result)
```
lcarcaramo, line 306:
```python
match = re.search(racf_command_argument_regex, security_result)
```

pyracf/common/utilities/logger.py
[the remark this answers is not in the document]
lcarcaramo: I think this is ok now. There is just a typo that I think needs to be fixed:

change
```
a'(' before the ')'.
```

to
```
a '(' before the ')'.
```

tests/user/test_user_result_parser.py
lcarcaramo: Just to confirm, redaction for extract is not within the scope of this pull request?

tests/group/test_group_debug_logging.py
```diff
@@ -35,7 +35,7 @@ def test_alter_group_request_debug_log_works_on_success(
         stdout = io.StringIO()
         with contextlib.redirect_stdout(stdout):
             self.group_admin.alter(
-                "TESTGRP0", traits=TestGroupConstants.TEST_ALTER_GROUP_REQUEST_TRAITS
+                "testgrp0", traits=TestGroupConstants.TEST_ALTER_GROUP_REQUEST_TRAITS
```
lcarcaramo, line 38: What is the nature of these changes?

